PR #5, opened 2015-04-20
Better sourcemaps support

badsyntax:
This fixes source mapping when used with sass-loader.

Also, we don't want to map to the postcss generated source, we want to map to the pre-postcss source, so this is also an improvement for CSS source maps.

Inline source maps can be generated using `devtool: 'inline-source-map'`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,13 +1,13 @@
 var loaderUtils = require('loader-utils');
 var postcss     = require('postcss');
 
-module.exports = function (source) {
+module.exports = function (source, map) {
     if ( this.cacheable ) this.cacheable();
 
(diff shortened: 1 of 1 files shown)


badsyntax:
Thanks for the merge. This change breaks the tests though, I'm wondering if it would be better to allow configuration of map options with loader params...

ai:
@badsyntax np, I fixing test right now. Also I will move loader to PostCSS 4.1.

badsyntax:
Great, thanks a lot :+1:

ai:
Released in 0.4.0 with all other PostCSS 0.4.1 features

badsyntax:
Thanks a lot!

PR #6, opened 2015-05-02
Fixed invalid sourcemap exception

badsyntax:
PostCSS was throwing "Unsupported previous source map format" error in situations where a sass file was compiled without any output. (For example a scss file that contained an empty selector.)

### DIFF
--- a/index.js
+++ b/index.js
@@ -15,7 +15,7 @@ module.exports = function (source, map) {
             annotation: false
         }
     };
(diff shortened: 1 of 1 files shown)


ai:
@badsyntax I will merge it when we will finish PostCSS Runner/Plugin Guidelines

badsyntax:
@ai :+1:

PR #82, opened 2016-07-18
Update README.md to fix example

bcbcb:
A comment in the example code refers to the CSSWring plugin but it should be PreCSS instead.


ai:
Thanks!

PR #87, opened 2016-08-15
Consistent format for CSS syntax errors

andreypopp:
I used `babel-code-frame` package formatting syntax errors code frames. That makes it consistent with Babel output which is I think very important as it makes it less cognitive efforts to recognize and process the error. Also code frames are a little bit fancies now (shows the line numebrs) so probably PostCSS itself should adopt this style.

You can also see that there's no duplicated filename in error message as webpack already print module name.

Before:

screenshot 2016-08-15 19 40 31

After:

screenshot 2016-08-15 19 39 19

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,31 @@
-var loaderUtils = require('loader-utils');
-var postcss     = require('postcss');
+var formatCodeFrame = require('babel-code-frame');
+var loaderUtils     = require('loader-utils');
+var postcss         = require('postcss');
+
+function formatMessage(message, loc, source) {
+    var formatted = message;
+    if (loc) {
+        formatted = formatted +
+            ' (' + loc.line + ':' + loc.column + ')';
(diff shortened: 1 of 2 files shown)


ai:
Awesome. It will be very useful.

But maybe we should change it in PostCSS core? Could you send PR there?

andreypopp:
Hey @ai, I'd be glad to make a PR to PostCSS core but this PR is specific to webpack as doesn't render filename of the file errors happened into (something you don't want to miss in general case). 

I'd suggest taking this PR here and then I'd make PostCSS reuse `babel-code-frame` for source code formatting.

ai:
Yeap:
1. We could add line numbers to PostCSS core.
2. Then we could fix file path in error in loader.

Looks good?

andreypopp:
Thanks @ai!

PR #93, opened 2016-08-24
Add postcss-loader-before-processing event

jantimon:
This pull request introduces a new feature.

It allows third party webpack plugins to add or remove postcss plugins using the `postcss-loader-before-processing` event (using the webpack built in plugin system).

An example can be seen here:
https://github.com/jantimon/postcss-loader/blob/33436d715119258053e5459b7cfde26eeecd0aa1/test/webpack-plugins/rewrite.js

Motivation:
There are cases where a postcss plugin is very tight cupeled with a webpack plugin as it has to access the webpack internals.
In this case it would be nice if this webpack could also add its postcss part to the loader.

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,4 +290,15 @@ export default {
 }
 ```
 
+
+### Webpack Events
+
+Webpack provides webpack-plugin developers a convenient way to hook into the build pipeline.
+The postcss-loader makes us of this event system to allow building integrated postcss-webpack tools.
+
+See the [example implementation](https://github.com/postcss/postcss-loader/blob/master/test/webpack-plugins/rewrite.js)
+
+* `postcss-loader-before-processing`  
+  is fired before processing and allows to add or remove postcss plugins
+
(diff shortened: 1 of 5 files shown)


ai:
Looks great. Could I ask you two changes:

1. Remove `ChangeLog` and `version` release commits. I will make separated config with version bump with GPG sign (I am cryptopunk :D ).
2. We need better docs for this feature (my English is poor). We need to describe how other developers could use this feature in their plugins.

jantimon:
Okay cool will update the pr tomorrow - I could also link to the rewrite example in the readme

jantimon:
Updated the pull request as requested. :)

ai:
Looks good. Wait I second, I need to finish my talk and them I will release all changes in PostCSS projects.

You don’t need this feature right now, correct?

jantimon:
Cool :) which talk will you give? releasing in the near future would be great but doesn't need to be right now.

ai:
Released in 0.11.

PR #95, opened 2016-08-31
Runaway promise

leyyinad:
In conjunction with bluebird, I get a couple of Warnings:

> `Warning: a promise was created in a handler at usr/app/node_modules/postcss-loader/index.js:100:13 but was not returned from it, see http://goo.gl/rRqMUw`

[Documentation](http://goo.gl/rRqMUw) states that one should at least `return null` to "signal that we didn't forget to return".


ai:
Thanks!

ai:
Released in 0.11.1

Issue #86, opened 2016-08-13
Workaround for this.exec?


7rulnik:
Hi! Thanks for postcss and loader for it. It's really awesome. But…

I want to use SugarSS parser with https://github.com/amireh/happypack but it cannot work with `this.exec`.
Does it really necessary? Or maybe exist any workarounds?

ai:
Hi. Thanks for good issue start text 😆

What you mean about `this.exec`? How it could be fixed? (I didn't use HappyPack).

7rulnik:
[Here](https://github.com/postcss/postcss-loader/blob/master/index.js#L59) you use `this.exec` but happypack [can't support](https://github.com/amireh/happypack/wiki/Webpack-Loader-API-Support) it.

UPD: oh, I was wrong. You use exec only for `postcss-js` parser.
Sorry T____T

ai:
So, how I could fix it?

PR #96, opened 2016-09-04
Allow control over loader `exec`.

nealgranger:
An `exec` flag can be provided to the loader query or webpack config. This instructs the loader to load the source using `exec`. This enhances the previous behavior which was to check that the loader query parser was strictly equal to `postcss-js`.

This fixes an issue where the `postcss-js` parser cannot be specified through the webpack postcss config option.

This also enables parsers other than `postcss-js` to receive executed module results.

### DIFF
--- a/index.js
+++ b/index.js
@@ -51,6 +51,7 @@ module.exports = function (source, map) {
     }
 
     var plugins;
+    var exec;
     if ( typeof options === 'undefined' ) {
         plugins = [];
     } else if ( Array.isArray(options) ) {
@@ -60,6 +61,7 @@ module.exports = function (source, map) {
         opts.stringifier = options.stringifier;
         opts.parser      = options.parser;
         opts.syntax      = options.syntax;
+        exec             = options.exec;
     }
     if ( params.pack ) {
         plugins = options[params.pack];
@@ -77,11 +79,14 @@ module.exports = function (source, map) {
     if ( params.stringifier ) {
         opts.stringifier = require(params.stringifier);
     }
+    if ( params.exec ) {
+        exec = params.exec;
+    }
 
     var loader   = this;
     var callback = this.async();
 
-    if ( params.parser === 'postcss-js' ) {
+    if ( params.parser === 'postcss-js' || exec ) {
         source = this.exec(source, this.resource);
     }
 
--- a/test/cases/exec.js
+++ b/test/cases/exec.js
@@ -1,0 +1,9 @@
+var postcssJs = require('postcss-js');
+
+var style = {
+    a: {
+        color: 'green'
+    }
+};
+
+module.exports = postcssJs.parse(style);
--- a/test/test.js
+++ b/test/test.js
@@ -31,6 +31,13 @@ describe('postcss-loader', function () {
         expect(css).to.eql('a {\n    color: red\n}');
     });
 
+    it('processes CSS with exec', function () {
+        var css = require('!raw-loader!' +
+                          '../?exec!' +
+                          './cases/exec.js');
+        expect(css).to.eql('a {\n    color: green\n}');
+    });
+
     it('inlines map', function () {
         var css = require('!raw-loader!../?sourceMap=inline!./cases/style.css');
         expect(css).to.include('/*# sourceMappingURL=');


ai:
Hm. Adding a new option looks like a hack. Other developers with same issue could miss this option.

Could you describe this issue, that you want to fix here? Maybe we will find some better way?

nealgranger:
This is basically a fix for #86

nealgranger:
It's useful to be able to conditionally load the required module with `exec`. This opens the door for parsers other than `postcss-js` and allows support for specifying the `postcss-js` parser in ways _other_ than `?parser=postcss-js` in the loader query.

We will be able to do this:
```js
module.exports = {
    module: {
        loaders: [
            {
                test:   /\.style.js$/,
                loader: "style-loader!css-loader!postcss-loader"
            }
        ]
    },
    postcss: function () {
        return {
            parser: require('postcss-js'),
            exec: true,
        };
    }
}
```

Or this:
```js
module.exports = {
    module: {
        loaders: [
            {
                test:   /\.style.xyz$/,
                loader: "style-loader!css-loader!postcss-loader?parser=custom-parser&exec"
            }
        ]
    },
}
```

nealgranger:
@ai Having the parser itself define the exec behaviour is also an option.

In https://github.com/postcss/postcss-js/blob/master/index.js, for example:

```diff
  var postcssJs = {
      objectify: require('./objectifier'),
      parse:     require('./parser'),
      async:     require('./async'),
      sync:      require('./sync'),
+     __webpacExec: true,
  };
  module.exports = postcssJs;
```

ai:
Yeap, we could change API. `postcss-js` is not so widely used, so we could change it.

ai:
I don’t like the `__webpackExec: true` property in parser. It looks like hack ;).

Maybe we should add `parserExec` parameter? But it will be a string parameter `?parserExec=postcss-js`.

I would be happy to avoid changing `array` to `object` in `postcss` section of webpack config (because it will affect to many users). Any suggestions?

nealgranger:
@ai What do you mean about "`array` to `object`". The changes I have staged are 100% compatible with the existing interface. This should have absolutely no impact on current users.

ai:
@nealgranger 

```js
postcss: function () {
        return {
            parser: require('postcss-js'),
            exec: true,
        };
    }
```

Right now `postcss` section in webpack returns only plugins list.

nealgranger:
there's already a case for this: https://github.com/postcss/postcss-loader/blob/master/index.js#L59. The webpack config can be an array or an object with a `plugins` array.

ai:
Ouh, I forget about it :(. Sure, we could add `parser` option to `postcss` section.

But what if developer need `postcss-js` to some files and CSS parser to others?

nealgranger:
@ai The loader query syntax is still supported of course.

I'm currently doing something similar to this:
```js
loaders: [
  {
    test:   /\.css$/,
    loader: "style-loader!css-loader!postcss-loader"
  },
  {
    test:   /\.css.js$/,
    loader: "style-loader!css-loader!postcss-loader?parser=postcss-js"
  }
]
```

ai:
I am suggesting to remove `postcss-loader?parser=postcss-js` for `postcss-js`.

Let’s introduce one way to load `exec` based parsers. Because I don’t like `exec` option idea — other developers could miss this option.

nealgranger:
@ai The `exec` option is not necessary if you are using `?parser=postcss-js`. This is a totally non-breaking change.

nealgranger:
@ai Were you thinking of a separate parameter like `?execParser=custom-parser`?

ai:
I know. But what if other developers got #86? How did he find this `exec` option?

This is why I suggest to change how we load `postcss-js`. So having #86 will be impossible.

nealgranger:
I understand #86 better now, `exec` is not supported when using `happypack`.

The solution would be to allow using plain old `require` instead.

nealgranger:
We would need a way to specify whether to use `loader.exec`, `require` or nothing at all.

ai:
Sure, let’s use something like this instead:

```js
    postcss: function () {
        return {
            parser: require('postcss-js'),
            plugins: [require('autoprefixer')]
        };
    }
```

ai:
Other question: how it will be changed in webpack 2 ;)

nealgranger:
..and have `?parser=postcss-js` automatically trigger `exec` for compatibility.

nealgranger:
how does
```
    postcss: function () {
        return {
            parser: require('postcss-js'),
            plugins: [require('autoprefixer')]
        };
    }
```
trigger using `exec`?

ai:
Hm, mayb I miss the point. Could you repeat what is the problem?

nealgranger:
How does the loader know to call `exec` in the example postcss config above?

ai:
@nealgranger I mean what is the main problem? #86 was that `exec` doesn’t work in some environments.

nealgranger:
I was wrong to say that this is a fix for #86

nealgranger:
To solve #86, we would need to do something like:

```js
if ( params.parser === 'postcss-js') {
  source = require(source);
}
```

which would prevent any code in the source file from being processed with webpack, but at least it would work.

ai:
OK. So why you need this PR? To run `postcss-js` behavior to some custom parsers?

nealgranger:
yup

nealgranger:
I'm currently doing:

```js
export const parse = (source) => postcssJs.parse(transformSource(source)));
```

nealgranger:
@ai It's getting used here https://github.com/webpack-config/webpack-config-postcss/compare/postcss-js?expand=1.

ai:
So you suggest `?parser=./my-parse& exec=true`. I think it will be better to replace it to `parserExec=my-parse`

ai:
Sure, let’s me finish with PostCSS 5.2 and I will return to this issue.

nealgranger:
I hadn't considered `parserExec=my-parse` until our conversation. I do prefer`?exec` though. It avoids the ambiguity of `?parserExec=my-parse?parser=other-parser`.

nealgranger:
@ai Thanks for looking at this!! 😄

ai:
Sure, `parserExec` is not a best way, because `exec` is about style file, to a parser.

ai:
Please check docs https://github.com/postcss/postcss-loader/commit/8388d3d705fa2b74ca0b37f260a5bb8ee7562ec8

ai:
Released in 0.13.

PR #101, opened 2016-09-21
Gracefully handle lack of compilation object

ertrzyiks:
That way loader will not be fully functional when used with happypack, but at least without a crash.

Before:
```
ERROR in ./main.css
Module build failed: Error: Cannot read property 'applyPluginsWaterfall' of undefined
TypeError: Cannot read property 'applyPluginsWaterfall' of undefined
```

After
```
WARNING in ./main.css
this._compilation is not available thus `postcss-loader-before-processing` is not supported
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -83,11 +83,18 @@ module.exports = function (source, map) {
     }
 
     // Allow plugins to add or remove postcss plugins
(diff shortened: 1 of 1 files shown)


ai:
Sorry, I will release it with config (maybe on next week)

PR #104, opened 2016-10-07
Allow `options` via `query`.

izaakschroeder:
When `webpack@2` lands there will no longer be any `this.options`. Since query can now be a vanilla object instead of a string, having `options` reside in query is a quick, non-destructive fix.

See:
 * https://github.com/webpack/extract-text-webpack-plugin/pull/266
 * https://github.com/webpack-config/webpack-config-postcss/pull/10

### DIFF
--- a/index.js
+++ b/index.js
@@ -37,7 +37,7 @@ module.exports = function (source, map) {
     if ( typeof map === 'string' ) map = JSON.parse(map);
     if ( map && map.mappings ) opts.map.prev = map;
 
-    var options = this.options.postcss;
+    var options = params.options || this.options.postcss;
     if ( typeof options === 'function' ) {
         options = options.call(this, this);
     }
--- a/index.js
+++ b/index.js
@@ -37,7 +37,7 @@ module.exports = function (source, map) {
     if ( typeof map === 'string' ) map = JSON.parse(map);
     if ( map && map.mappings ) opts.map.prev = map;
 
-    var options = params.options || this.options.postcss;
+    var options = params.plugins || this.options.postcss;
     if ( typeof options === 'function' ) {
         options = options.call(this, this);
     }


ai:
Why extra option instead of having option in root level in query?

izaakschroeder:
The `query` itself must either be `object` or `string` and it's nice to have `options` as a function. It's also the least intrusive change.

ai:
Maybe we should name it as `plugins`?

izaakschroeder:
Yea we could do `query.plugins` instead of that works. 👍

izaakschroeder:
Thanks! Let me know when a new version is publish @ai 😄

ai:
I just need to update docs, add result.messages with dependencies and we are ready for release.

izaakschroeder:
👌

ai:
I finished all changes and ready for release. Could you test this plugin on real application? (Just replace version in `package.json` to `postcss/postcss-loader`)

michael-ciniawsky:
webpack v2.1.0-beta.25

LoaderOptionsPlugin: ✅ 
postcss.config.js:  :no_entry_sign:

I'm on it... :)

izaakschroeder:
Will try and get this working will my configuration tonight 😄

izaakschroeder:
@ai Running into a few hiccups. First of which is ` ./node_modules/postcss-load-config/index.js`. There's some mismatch between versions going on here, where that file is trying to require `postcss-load-options/lib/loadOptions.js` which doesn't exist; the file is instead called `postcss-load-options/lib/options.js`. I'm not sure which version is intended to be correct or where the mismatch is coming from.

izaakschroeder:
@ai The way you've refactored `parseOptions` also results in failing to pass the webpack context through.

```js
return parseOptions(options, pack);
```

should be:

```js
var context = this; // at the top of the loader

// ...
return parseOptions(context, options, pack);
```

and then:

```js
function parseOptions(context, options, pack) {
// ...
options = options.call(context, context);
```

izaakschroeder:
You can verify these by checking out the branch here: https://github.com/webpack-config/webpack-config-postcss/pull/10 and running `npm run smoke`.

michael-ciniawsky:
@izaakschroeder postcss-load-config issue should be fixed now.

izaakschroeder:
Thanks @michael-ciniawsky it is 😄 Just waiting on @ai now.

PR #121, opened 2016-10-19
Fix wrong notation for loaders option on README

brnrdog:
Some examples was using the option loader for an array of loaders, but the option should be on plural when its value is an array. 

This PR fixes that.


ai:
Thanks, it is very big docs mistake. It is good that you find it :+1:

PR #126, opened 2016-10-28
update readme config section - loaders order

code4aman:
Hi, fixed a bit confusing moment ^_^


michael-ciniawsky:
Yep 👍  that's definitely confusing, maybe it is the 'webpack' way to describe it or simply a typo ¯\_(ツ)_/¯, always a bit messy 😛 . Thx

PR #140, opened 2016-11-15
Update minimum NodeJS and NPM versions required

ntwb:
This module requires `postcss-load-config` which in turn requires `postcss-load-plugins` and `postcss-load-options`, all three of these repos have a NodeJS & NPM minimum requirement:

• "engines": { "node": ">=0.12", "npm": ">=3" },`

https://github.com/michael-ciniawsky/postcss-load-config/blob/master/package.json#L5
https://github.com/michael-ciniawsky/postcss-load-plugins/blob/master/package.json#L5
https://github.com/michael-ciniawsky/postcss-load-options/blob/master/package.json#L5

To prevent warnings when using this module this repo should also follow the same min req.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "postcss-loader",
   "version": "1.1.1",
   "description": "PostCSS loader for webpack",
+  "engines": { "node": ">=0.12", "npm": ">=3" },
   "keywords": ["webpack", "loader", "css", "postcss", "postcss-runner"],
   "author": "Andrey Sitnik <andrey@sitnik.ru>",
   "license": "MIT",


ai:
@ntwb why we should add `engines` here and not to `postcss-load-config`?

michael-ciniawsky:
@ai Does hurt to provide metadata about node and npm version, I would say merge it 😛 

@ntwb Does`npm install --save|--save-dev postcss-loader` currently display any warnings, because of missing pkg.engines prop?

ntwb:
@ai, `postcss-load-config` already has `engines` defined: https://github.com/michael-ciniawsky/postcss-load-config/blob/master/package.json#L5

Here's the details of Travis CI job I noticed this in:  https://travis-ci.org/stylelint/stylelint-demo/jobs/175820550#L172

ai:
@ntwb so, do we need a `engines` in `postcss-loader` if it is already in `postcss-load-config`?

ntwb:
> @ntwb so, do we need a engines in postcss-loader if it is already in postcss-load-config?

Yes, which is what this PR fixes 😄

ai:
@ntwb and why we need it if it is already in `postcss-load-config`?

ntwb:
If a repo you depend upon requires NodeJS v7.x, then this repo should also state that the minimum requirement is NodeJS v7.x

In this case though `postcss-load-config` has a minimum NPM requirement for NPM v3.x, so that same minimum requirement should be included in `postcss-loader`

ai:
I mean when benefits user will get? npm warning will be the same?

ntwb:
They will no longer get the NPM warning

ntwb:
Thanks :)

PR #142, opened 2016-11-22
fix wrong extension on example code

astroplex:


ai:
It is correct extension for SugarSS 😊 this example show how to change a parser.

But thanks for be attentive.

PR #143, opened 2016-12-02
You can now set custom config path.

sainthkh:
I personally organize all of my build settings to "build" folder. However, postcss-loader only accepts configuration files in its ancestor folders. 

So, I changed the code to accept the custom configuration path by 
`'postcss?config=./the/path/to/your/config/file/from/the/root',`

 I hope you like it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -48,6 +48,7 @@ module.exports = function (source, map) {
     var options  = params.plugins || loader.options.postcss;
     var pack     = params.pack;
     var callback = loader.async();
+    var config_path = params.config ? path.join(process.cwd(), params.config) : path.dirname(file)
 
     Promise.resolve().then(function () {
         if ( typeof options !== 'undefined' ) {
@@ -57,7 +58,7 @@ module.exports = function (source, map) {
                 throw new Error('PostCSS plugin pack is supported ' +
                                 'only when use plugins in webpack config');
             }
-            return loadConfig({ webpack: loader }, path.dirname(file));
+            return loadConfig({ webpack: loader }, config_path);
         }
     }).then(function (config) {
         if ( !config ) config = { };


ai:
Thanks for PR. Let’s discuss one thing.

Main idea behind common PostCSS config is to share same config between different tools: webpack-loader, IDE, etc. With `config` path other tools will not find config and will work in different way.

Does `babel` CLI has option for config? Where you put ` .babelrc`?

sainthkh:
As for your question, it seems that you cannot change your `.babelrc` path.

Actually, I don't use `.babelrc` file and I included the babel options in my webpack.config.js. In there, there are css-loader for CSS Modules and postcss-loader for those CSS files. The loader setting is like below.

```
loaders: [{  
	test: path.join(__dirname, '..', 'app'),  
	loader: ['babel-loader'],  
	query: {
		cacheDirectory: './temp/babel_cache',
		presets: ['react', 'es2015']
	}
}, {
	test: /\.css$/,
	loaders: [
		'style?sourceMap',
		'css?modules&importLoaders=1&localIdentName=[path]___[name]__[local]___[hash:base64:5]',
		'postcss?config=./build/postcss.config.js',
	]
}]
```

And below is my project folder.
```
+app
- Various Components and code
+build
-webpack.config.js
-pocstcss.config.js
+static
-various static assets
index.js
package.json
```

There are some people like me [who love organizing config files in one place.](https://blog.risingstack.com/node-hero-node-js-project-structure-tutorial/) I think it's not a harmful option. 

As for webpack-dev-server, it has `--config` option, too. So, it works perfect for the people like me. 

p.s. 
I checked the travis CI and it seems that I ignored the coding convention. If you decide to accept this PR, I will change that code to meet the convention.

ai:
I like clean project tree too. But I think right now root is a common place for config 😆.

Because I love global config idea. What do you think about this solution. We will add this option, but still not as part of public API?

It will not be in docs to not blocking global config promotion. OK?

sainthkh:
Good. 

I agree that my personal preference(separate folder for config) is a bit minor. So, I think that your solution is the best for everyone. 

Then, I will send you another PR that meets code convention.

michael-ciniawsky:
@sainthkh The behaviour you want to achieve is having all config files in a separate  e.g `./config` folder? Because `postcss.config.js` should be loaded relative to the current file processed by `postcss-loader` e.g

```
|–src
|   |–styles
|       |–index.css
|       |–postcss.config.js
|   |–scripts
|       |–index.js
|       |–.babelrc
|   |–components
|       |–index.html
|       |–posthtml.config.js
|–webpack.config.js / gulpfile.js / gruntfile.js / postcss -i ... -o result.css (CLI)
|–package.json
```

If you want to have it all in one place and don't need a function as config => `package.json`

**package.json**
```js
{
  "name": "",
   "version": "",
    "scripts": "",
   "dependencies": {}
   "postcss": { parser:  "",  plugins: {}}
   "babel":  { env: {}, presets: [] }
    ...
}
```

sainthkh:
@michael-ciniawsky Yes, that's possible. However, PostCSS has a lot of configuration. In json file, you cannot handle them. 

So, if you want to move postcss.config.js to `./config` and handle the configuration, the code change is necessary.

michael-ciniawsky:
> So, if you want to move postcss.config.js to ./config and handle the configuration, the code change is necessary.

@sainthkh Yep, I understand, was not meant to be against it, asked to clearify usecase :)

PR #147, opened 2016-12-12
Allow absolute path as config option

SimenB:
I'm unable to write a test for it as you compile the tests, so webpack handles the path for you... Ideas?

The test I wrote:
```js
it('allows to change config path with absolute path', function () {
    // eslint-disable-next-line no-path-concat
    var css = require('!raw-loader' +
                      '!../?config=' +
                      __dirname + '/cases/config/postcss.config.js' +
                      '!./cases/style.css');
    expect(css).toEqual('a { color: black }\n');
});
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -48,8 +48,17 @@ module.exports = function (source, map) {
     var options  = params.plugins || loader.options.postcss;
     var pack     = params.pack;
     var callback = loader.async();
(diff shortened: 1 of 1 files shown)


ai:
Thanks! Release in 1.2.1.

PR #154, opened 2016-12-21
Adding description of using postcss-js parser with babel 6 compiler

komlev:
postcss-js parser produce wrapped styles, adding description to fix this problem

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,10 +180,10 @@ export default {
 }
 ```
 
-> If you are using babel 6 compiler and up you need to do the following in order for setup to work
+> If you are using Babel >= v6 you need to do the following in order for the setup to work
 
-> * It is suggested to have only **default** export per style module
-> * You need to add [babel-plugin-add-module-exports] to you configuration
+> 1. Add [babel-plugin-add-module-exports] to your configuration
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -180,6 +180,11 @@ export default {
 }
 ```
 
+> If you are using babel 6 compiler and up you need to do the following in order for setup to work
```
michael-ciniawsky: - `babel 6 compiler` => `Babel >= v6` 
- in order for setup  => in order for *the* setup to work

README.md
```diff
@@ -180,6 +180,11 @@ export default {
...
+
+> * It is suggested to have only **default** export per style module
```
michael-ciniawsky: only => only *one*  || only => one ?

README.md
```diff
@@ -180,6 +180,11 @@ export default {
...
+> * You need to add [babel-plugin-add-module-exports] to you configuration
```
michael-ciniawsky: Maybe switch order here, but I'm nitpicking 🐤 here 😛 

1. add babel-plugin-add-module-exports
2. only one **default** export per module/file

komlev:
Done

PR #171, opened 2017-02-10
Don't force internal requires of options

jescalan:
Hi there! So I ran into a situation where I need to require the parser myself and pass it in, rather than having it required internally, so I added this option. I couldn't quickly get my head around how the test suite works, but if you give me a push in the right direction I'd be happy to add a test for this. Meanwhile, it is a progressive upgrade, so it won't break any existing functionality.


ai:
Good change. I will accept it and release on this weekend (sorry, I am in the travel right now).

ai:
Released in 1.3.0.

PR #174, opened 2017-02-18
Do not load config file from `--config`

egoist:
Because no need to and some other tools might use `--config` to load their config file.


ai:
Thanks. Released in 1.3.1.

michael-ciniawsky:
@egoist Thx 😛

PR #180, opened 2017-03-01
Remove webpack Plugin warning (#179)

### LINKED ISSUE #179: Allow to suppress warning when used with happypack
[There](https://github.com/postcss/postcss-loader/blob/master/index.js#L128) is a warning, which warns happypack user about lack of support for `postcss-loader-before-processing` event to prevent potential failure with no clear reason.

Once the problem is acknowledged / resolved on the user side, we may not need that warning in the build output anymore. How could we achieve that?

My ideas:
 - introduce an option to explicitly disable `postcss-loader-before-processing` event
 - introduce an option to ignore this particular warning

If you are happy about any of those solutions or know better one, I'm willing to help to make it happen.

michael-ciniawsky:
Fixes #179

### DIFF
--- a/index.js
+++ b/index.js
@@ -124,11 +124,6 @@ module.exports = function (source, map) {
               [].concat(plugins),
               params
             );
-        } else {
-            loader.emitWarning(
-              'this._compilation is not available thus ' +
-              '`postcss-loader-before-processing` is not supported'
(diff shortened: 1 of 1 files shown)


ai:
Released in 1.3.3

PR #186, opened 2017-03-14
doc(): update webpack v2 part

kud:


mrmartineau:
@kud I was going to update this myself. Do you think that an example with the webpack extract-text plugin should also be included: https://github.com/postcss/postcss/pull/989/files#diff-04c6e90faac2675aa89e2176d2eec7d8R214

kud:
Could be interesting indeed. :)

kud:
I used this conf on another project, yes.

PR #201, opened 2017-04-23
fix: return promise to correctly resolve options after first call

nsfmc:
when using inlined plugin options (not using `postcss.config.js`) the
first call to the loader would succeed in finding any `options` set,
but subsequent calls would fail.

the fix here changes the loader to return the promise it creates which
seems to allow using an inlined `options.plugins` function again.

### DIFF
--- a/index.js
+++ b/index.js
@@ -61,7 +61,7 @@ module.exports = function (source, map) {
         configPath = path.dirname(file);
     }
 
-    Promise.resolve().then(function () {
+    return Promise.resolve(function () {
         if ( typeof options !== 'undefined' ) {
             return parseOptions.call(loader, options, pack);
         } else {


michael-ciniawsky:
@nsfmc Some tests are failing I take a deeper look tomorrow 😛

nsfmc:
fwiw, i see those tests failing, but i tested with this project https://github.com/nsfmc/extract-text-test and i see @value imports working as well as composing doing the right thing, clearly not an exhaustive test or anything, but a not bad smoketest if you're looking for a bare repo to test against.

nsfmc:
hahahaha, i want to apologize @michael-ciniawsky , i totally did not understand how loaders work at all when i opened this pr (and still my understanding of the pattern they use for async execution is hazy). this pr does _something_ but it's not correct... _at all_, however, if you take a look at https://github.com/nsfmc/extract-text-test and revert the `postcss-loader` req back to `^1.3.3` you'll see that the problem does seem to come from postcss loader assuming that it will get its config from the `postcss.config.js` file

**build failure when running `yarn build-plain`**
<details>
```shell
marcos@Dalarna in ~/projects/extract-text-test on master!
± yarn build-plain
yarn build-plain v0.23.2
$ webpack --config regular-postcss-config.js 
Hash: b9d02904b81cc385128c
Version: webpack 2.4.1
Time: 504ms
   Asset     Size  Chunks             Chunk Names
index.js  74.1 kB       0  [emitted]  main
   [0] ./~/css-loader?{"modules":true,"localIdentName":"[name]__[local]___[hash:base64:5]","importLoaders":1}!./~/postcss-loader?{}!./src/index.css 240 bytes {0} [built] [failed] [1 error]
   [1] ./src/styles.css 1.08 kB {0} [built]
   [2] ./~/base64-js/index.js 3.48 kB {0} [built]
   [3] ./~/buffer/index.js 48.6 kB {0} [built]
   [4] ./~/css-loader?{"modules":true,"localIdentName":"[name]__[local]___[hash:base64:5]","importLoaders":1}!./~/postcss-loader?{}!./src/styles.css 929 bytes {0} [built]
   [5] ./~/css-loader/lib/css-base.js 2.19 kB {0} [built]
   [6] ./~/ieee754/index.js 2.05 kB {0} [built]
   [7] ./~/isarray/index.js 132 bytes {0} [built]
   [8] ./~/style-loader/addStyles.js 8.51 kB {0} [built]
   [9] ./~/style-loader/fixUrls.js 3.01 kB {0} [built]
  [10] (webpack)/buildin/global.js 509 bytes {0} [built]
  [11] ./src/index.js 51 bytes {0} [built]

ERROR in ./~/css-loader?{"modules":true,"localIdentName":"[name]__[local]___[hash:base64:5]","importLoaders":1}!./~/postcss-loader?{}!./src/index.css
Module build failed: Error: No PostCSS Config found in: /Users/marcos/projects/extract-text-test/src
    at Error (native)
    at /Users/marcos/projects/extract-text-test/node_modules/postcss-load-config/index.js:51:26
 @ ./~/css-loader?{"modules":true,"localIdentName":"[name]__[local]___[hash:base64:5]","importLoaders":1}!./~/postcss-loader?{}!./src/styles.css 3:10-209 6:63-262 10:13-212
 @ ./src/styles.css
 @ ./src/index.js
error Command failed with exit code 2.
```
</details>

when i console.log this, i notice that the first execution of the loader, it picks up its options correctly, but i _think_ that subsequently, the loader runner is not using the previous config (and i still don't understand enough to know why other than to think that it is looking for config in the webapp scope and finding nothing and bailing without looking for it in the original loader def).

again, apologies for this wild goose chase, i'm trying to figure out more, but this isn't top of mind for me, so i'm just picking at it for a few minutes each day.

michael-ciniawsky:
@nsfmc Sry for the delay I've been busy 😛 , the config in your example looks sane and when postcss options are set in `webpack.config.js`, `postcss-load-config` (`postcss.config.js`) shouldn't trigger at all, I will test the  upcoming `postcss-loader` v2.0.0 with ETWP extensively and marked your example to do so :) Meanwhile I close this PR since it isn't fixing the issue and I'm on the finishing line for v2.0.0 💯 

Tracking issue is #205, so if you find anything that could be a possible hint please let me know 🙃 

Thx 👍

nsfmc:
thanks for taking a look @michael-ciniawsky sorry i didn't have time to figure out how to fix the problem though

michael-ciniawsky:
It's complicated and could be a bug somewhere else enterily, see #203, the cause in this issue is unrelated to the problem here enterily, but some error happend elsewhere in `css-loader` and then the options in `postcss-loader` set in `webpack.config.js` did get picked up anymore ¯\_(ツ)_/¯

Issue #207, opened 2017-05-08
[2.0.1] "Cannot create property 'prev' on boolean 'false'" (options.sourceMap) 


le0nik:
```
Module build failed: TypeError: Cannot create property 'prev' on boolean 'false'
    at Promise.resolve.then.then (../node_modules/postcss-loader/lib/index.js:125:47)
```

```
    {
      loader: 'postcss-loader',
      options: {
        sourceMap: true,
        ...
      },
    },

```

These lines are the culprit:
https://github.com/postcss/postcss-loader/blob/master/lib/index.js#L93
https://github.com/postcss/postcss-loader/blob/master/lib/index.js#L125

postcss@1.3.3 works fine

webpack@2.5.1
OSX 10.12.4
Node.js 7.10.0
postcss-loader@2.0.0

michael-ciniawsky:
Thats weird 😛 , you don't need to set `sourceMap: false`  no source map is the default, still needs to be fixed I wonder why this happens tbh

polco:
i got the same error with the sourceMap set to 'true'  ;)

michael-ciniawsky:
hmm.... 😛 on it

le0nik:
@polco oh sorry my fault, this error appears when it's set to `true`, not `false`

michael-ciniawsky:
@le0nik @polco Please post your config (`webpack.config.js`)  aswell just in case :)

le0nik:
@michael-ciniawsky my config is huge and complex. Here's the styles loaders part:

```
    {
      loader: 'style-loader',
      options: {
        sourceMap: env.isDev,
      },
    }
    {
      loader: 'css-loader',
      options: {
        modules: true,
        importLoaders: 4,
        localIdentName: '[local]--[hash:base64:8]',
        minimize: {
          options: {
            safe: true,
            sourcemap: env.isDev,
          },
          // Otherwise it changes order of `z-index` which breaks our styles
          zindex: false,
        },
        sourceMap: env.isDev,
      },
    },
    {
      loader: 'resolve-url-loader',
      options: {
        sourceMap: env.isDev,
      },
    },
    {
      loader: 'postcss-loader',
      options: {
        sourceMap: env.isDev,
        plugins: getPostcssPlugins,
      },
    },
    {
      loader: 'sass-loader',
      options: {
        sourceMap: env.isDev,
      },
    },
```

The error happened when `env.isDev` was set to `true`.

[...]

PR #208, opened 2017-05-08
fix(index): do not delete options

frederikschubert:
The deletion of the options object causes
them to be undefined for subsequent uses
of the loader.

Closes https://github.com/postcss/postcss-loader/issues/207

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -69,11 +69,7 @@ module.exports = function loader (css, map) {
       : rc.ctx.options = {}
   }
 
-  delete options.config
-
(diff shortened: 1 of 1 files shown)


frederikschubert:
This also fixes the issue that the `postcss.config.js` can not be found when setting the path via the `options.config.path` property.

michael-ciniawsky:
Thx for quick PR 😛 , but already done in https://github.com/postcss/postcss-loader/commit/c4f006472a85d3bc36ae3f2f0e750291fd732b7b

Issue #217, opened 2017-05-09
[2.0.2] Don't fail on `sourceMap: false` && previous map (options.sourceMap)


doberkofler:
After upgrading to postcss-loader 2.0.2 I received the following type of error message for almost all files:
```
Module build failed: ModuleBuildError: Module build failed: TypeError: Cannot create property 'prev' on boolean 'false'
    at Promise.resolve.then.then (D:\MyDev\ljs_app\trunk\periscope\build\node_modules\postcss-loader\lib\index.js:131:47)
    at runLoaders (D:\MyDev\ljs_app\trunk\periscope\build\node_modules\webpack\lib\NormalModule.js:192:19)
    at D:\MyDev\ljs_app\trunk\periscope\build\node_modules\loader-runner\lib\LoaderRunner.js:364:11
    at D:\MyDev\ljs_app\trunk\periscope\build\node_modules\loader-runner\lib\LoaderRunner.js:230:18
    at context.callback (D:\MyDev\ljs_app\trunk\periscope\build\node_modules\loader-runner\lib\LoaderRunner.js:111:13)
    at Promise.resolve.then.then.catch (D:\MyDev\ljs_app\trunk\periscope\build\node_modules\postcss-loader\lib\index.js:172:71)
```

michael-ciniawsky:
Did you set { sourceMap: true } in `postcss-loader` when using source maps? It fails because `postcss-loader` doesn't generate a source map by default anymore and dicardes the previous source map, albeit the hard fail is a bug 😛

doberkofler:
@michael-ciniawsky Thank you. It's working now.

michael-ciniawsky:
Fixed by https://github.com/postcss/postcss-loader/commit/159b66a
Released in [v2.0.3](https://github.com/postcss/postcss-loader/releases/tag/v2.0.3)

duleigiser:
I  use **ExtractTextPlugin** to put all kinds of style files into one files.
> here is my config files fragments
```
{
        test: /\.(css$|styl)$/,
        use: ExtractTextPlugin.extract({
          fallback: 'style-loader',
          use://"css-loader!postcss-loader!stylus-loader"
           [
             "css-loader",
           {
              loader:"postcss-loader",
              options: {
               sourceMap:true //webpack config when we set `dev-tool: true'
             }
            },
            {
              loader:"stylus-loader",//put this loader  in the end
              options: {
                importLoaders: 1
              }
            }
          ]
        })
      }

```

[...]

PR #220, opened 2017-05-09
fix: fail on `sourceMap: false` && previous map (`options.sourceMap`)

jerexyz:
https://github.com/postcss/postcss-loader/issues/217


michael-ciniawsky:
@jerexyz 👍 That could fix it, but could you make this changes agianst current master this is the `v1.0.0` which doesn't have this issue :)

michael-ciniawsky:
Fixed && superseded  by https://github.com/postcss/postcss-loader/commit/159b66a Thx 😛

PR #229, opened 2017-05-10
fix: regression with `options.plugins` `{Function}` (`webpack.config.js`)

evilebottnawi:


michael-ciniawsky:
https://github.com/postcss/postcss-loader/releases/tag/v2.0.5

PR #234, opened 2017-05-16
refactor: improved loading algorithm of options and configs

evilebottnawi:
Why?
1. No todo :smile:
2. Always load config if your pass `config.path`, now it is not working as expected if i pass `config.path` and `sourceMap: false` my config file not loading -> `options.config && !sourceMap && length > 1` (i want to load config and set `sourceMap` use loader options ), also sometimes i want setup some option from options loader.
3. Some perf (we don't create `rc` object because it is not necessary if i use only `options` without `options.config`)
4. Allow to passing options to `postcss` throw `loader.options`, now it is possibly only using `postcss` config.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -47,53 +47,55 @@ module.exports = function loader (css, map) {
 
   validateOptions(require('./options.json'), options, 'PostCSS Loader')
 
-  const rc = {
-    path: path.dirname(file),
-    ctx: {
-      file: {
-        extname: path.extname(file),
-        dirname: path.dirname(file),
-        basename: path.basename(file)
-      },
-      options: {}
-    }
-  }
-
-  if (options.config) {
-    if (options.config.path) {
-      rc.path = path.resolve(options.config.path)
-    }
-
-    if (options.config.ctx) {
-      rc.ctx.options = options.config.ctx
-    }
-  }
-
-  const sourceMap = options.sourceMap
-
   Promise.resolve().then(() => {
-    const length = Object.keys(options).length
+    if (options.exec || options.parser || options.syntax || options.stringifier || options.plugins) {
+      return parseOptions.call(this, options)
+    }
 
-    // TODO
-    // Refactor
-    if (!options.config && !sourceMap && length) {
-      return parseOptions.call(this, options)
-    } else if (options.config && !sourceMap && length > 1) {
-      return parseOptions.call(this, options)
-    } else if (!options.config && sourceMap && length > 1) {
-      return parseOptions.call(this, options)
-    } else if (options.config && sourceMap && length > 2) {
-      return parseOptions.call(this, options)
+    const rc = {
+      path: path.dirname(file),
+      ctx: {
+        file: {
+          extname: path.extname(file),
+          dirname: path.dirname(file),
+          basename: path.basename(file)
+        },
+        options: {}
+      }
+    }
+
+    // Read options from options.config only when options.config is object
+    if (options.config) {
+      if (options.config.path) {
+        rc.path = path.resolve(options.config.path)
+      }
+
+      if (options.config.ctx) {
+        rc.ctx.options = options.config.ctx
+      }
     }
 
     return postcssrc(rc.ctx, rc.path, { argv: false })
-  }).then((config) => {
-    if (!config) config = {}
+      .then((config) => {
+        // Get `sourceMap` option from loader options when no `map` option in `postcss.config.js`
+        if (!config.options.map) config.options.map = options.sourceMap
 
+        return config
+      })
+  }).then((config) => {
     if (config.file) this.addDependency(config.file)
 
+    let sourceMap = config.options.map
     let plugins = config.plugins || []
-    let options = Object.assign({
+
+    // Disable override `to` option
+    if (config.options.to) delete config.options.to
+    // Disable override `from` option
+    if (config.options.from) delete config.options.from
+    // Disable override `map` option
+    if (config.options.map) delete config.options.map
+
+    let postcssOption = Object.assign({
       to: file,
       from: file,
       map: sourceMap
@@ -105,20 +107,20 @@ module.exports = function loader (css, map) {
 
     // Loader Exec (Deprecated)
     // https://webpack.js.org/api/loaders/#deprecated-context-properties
-    if (options.parser === 'postcss-js') {
+    if (postcssOption.parser === 'postcss-js') {
       css = this.exec(css, this.resource)
     }
 
-    if (typeof options.parser === 'string') {
-      options.parser = require(options.parser)
+    if (typeof postcssOption.parser === 'string') {
+      postcssOption.parser = require(postcssOption.parser)
     }
 
-    if (typeof options.syntax === 'string') {
-      options.syntax = require(options.syntax)
+    if (typeof postcssOption.syntax === 'string') {
+      postcssOption.syntax = require(postcssOption.syntax)
     }
 
-    if (typeof options.stringifier === 'string') {
-      options.stringifier = require(options.stringifier)
+    if (typeof postcssOption.stringifier === 'string') {
+      postcssOption.stringifier = require(postcssOption.stringifier)
     }
 
     // Loader API Exec (Deprecated)
@@ -132,10 +134,10 @@ module.exports = function loader (css, map) {
     }
 
     if (sourceMap && typeof map === 'string') map = JSON.parse(map)
-    if (sourceMap && map) options.map.prev = map
+    if (sourceMap && map) postcssOption.map.prev = map
 
     return postcss(plugins)
-      .process(css, options)
+      .process(css, postcssOption)
       .then((result) => {
         result.warnings().forEach((msg) => this.emitWarning(msg.toString()))
 
--- a/lib/options.js
+++ b/lib/options.js
@@ -11,12 +11,11 @@ module.exports = function parseOptions (params) {
   else if (Array.isArray(params.plugins)) plugins = params.plugins
   else plugins = params.plugins
 
-  const options = {}
-
-  if (typeof params !== 'undefined') {
-    options.parser = params.parser
-    options.syntax = params.syntax
-    options.stringifier = params.stringifier
+  const options = {
+    parser: params.parser,
+    syntax: params.syntax,
+    stringifier: params.stringifier,
+    map: params.sourceMap
   }
 
   const exec = params && params.exec
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "docs": "jsdoc2md lib/index.js > LOADER.md",
     "pretest": "npm run lint && npm run test:build",
     "test": "jest",
+    "test-only": "npm run test:build && jest && npm run clean",
     "test:build": "node test/webpack.build.js",
     "posttest": "npm run clean",
     "release": "standard-version"


lib/index.js
```diff
@@ -47,78 +47,67 @@ module.exports = function loader (css, map) {
...
-    }
 
     return postcssrc(rc.ctx, rc.path, { argv: false })
   }).then((config) => {
     if (!config) config = {}
 
     if (config.file) this.addDependency(config.file)
 
+    const sourceMap = options.sourceMap
+    delete options.sourceMap
```
michael-ciniawsky: This caused problems on 'multiple' runs for some weird reason, but maybe because it was defined in the wrong place, not sure here tbh ¯\_(ツ)_/¯. It work for the first loader run and was then missing for all further runs, which lead to `const sourceMap` being always `undefined`. Besides that `delete options.sourceMap` => `options.sourceMap = null` (perf) ?

lib/index.js
```diff
@@ -47,78 +47,67 @@ module.exports = function loader (css, map) {
...
-    path: path.dirname(file),
-    ctx: {
-      file: {
-        extname: path.extname(file),
-        dirname: path.dirname(file),
-        basename: path.basename(file)
-      },
-      options: {}
+  Promise.resolve().then(() => {
+    if (!options.config || (options.config && !options.config.path)) {
```
michael-ciniawsky: This changes to config loading behaviour ? See comment below where `options` get reassigned

package.json
```diff
@@ -33,6 +33,7 @@
     "docs": "jsdoc2md lib/index.js > LOADER.md",
     "pretest": "npm run lint && npm run test:build",
     "test": "jest",
+    "test-only": "npm run test:build && jest",
```
evilebottnawi: @michael-ciniawsky when i write code i want run only test to avoid regression (without lint), full test with lint and test i do before PR :smile:

lib/index.js
evilebottnawi: @michael-ciniawsky we don't load postcss config file when `config` not present or present but without `path`.

lib/index.js
evilebottnawi: @michael-ciniawsky weird, should works, I will test in the near future

lib/index.js
```diff
@@ -47,78 +47,67 @@ module.exports = function loader (css, map) {
...
+    let resolvedOptions = Object.assign({
       to: file,
       from: file,
       map: sourceMap
         ? sourceMap === 'inline'
           ? { inline: true, annotation: false }
           : { inline: false, annotation: false }
         : false
-    }, config.options)
+    }, options, config.options)
```
evilebottnawi: @michael-ciniawsky using `options` allow change `map`, it is bad, for some it is misleading behavior, but `to` and `from` now also we can change through `config.options`. I think about this.

lib/index.js
michael-ciniawsky: `postcss-load-config` has it's config lookup algo backed in via cosmiconfig, `options.config.path` should only be used for cases like 

```
|– .config
| |– postcss.config.js
|– src
| |– style.css
|
|– webpack.config.js
```

The current trigger for loading `postcss.config.js` is if no other options are set in `webpack.config.js` =>`postcss.config.js`

lib/index.js
michael-ciniawsky: > using options allow change map, it is bad, for some it is misleading behavior

? 🙃  You mean that currently one has to set `map` via `{ sourceMap: true || 'inline' }` (`webpack.config.js`) exclusively?

>  but to and from now also we can change through config.options

They shouldn't be changeable, since File I/O is handled by the webpack (Runner) 😛

lib/index.js
evilebottnawi: @michael-ciniawsky current behavior not works as expected we should fix this, see https://github.com/postcss/postcss-loader/pull/234#issue-229052711 (point 2).

evilebottnawi: @michael-ciniawsky two solution:
1. Load `postcss` config always and them merge option from loader with config options (i think loader options more important than `postcss` config option, but ).
2. Load `postcss` config only when `config: true` or `config: { path: '/path/to/config'}` (best way, no problems with perf and understandable using)

evilebottnawi:
@michael-ciniawsky we should finish and merge this PR asap, i spend about 5 hours for migrate to new postcss loader in most of my projects, in half the configuration is not loaded or source maps not working, current loading algorithm is misleading and incorrect.

My solution:
1. By default `config` is false and don't load postcss config.
2. If your have `config: true` or `config: { ctx : {...options}}` we loading config (also working good for nested config). Now `config: true` is disable by schema, but we can fix this.
3. If your have `config: { path: '/path/to/config' }` we loading config (absolutly).

This may seem small breaking changed, but it is bugfix, because now postcss config don't load in most use case.

lib/index.js
```diff
@@ -47,53 +47,55 @@ module.exports = function loader (css, map) {
...
+    if (!config.options) config.options = {}
 
     if (config.file) this.addDependency(config.file)
 
+    let sourceMap = options.sourceMap || config.options.map
     let plugins = config.plugins || []
-    let options = Object.assign({
+
+    // Disable override `to` option
+    if (config.options.to) delete config.options.to
```
evilebottnawi: Disable override important options from postcss `config`. it is break build.

lib/options.js
```diff
@@ -11,12 +11,11 @@ module.exports = function parseOptions (params) {
   else if (Array.isArray(params.plugins)) plugins = params.plugins
   else plugins = params.plugins
 
-  const options = {}
-
-  if (typeof params !== 'undefined') {
-    options.parser = params.parser
-    options.syntax = params.syntax
-    options.stringifier = params.stringifier
+  const options = {
```
evilebottnawi: We always have params as object, because we have `const options = loaderUtils.getOptions(this) || {}`

lib/index.js
```diff
@@ -47,53 +47,55 @@ module.exports = function loader (css, map) {
...
-    } else if (options.config && !sourceMap && length > 1) {
-      return parseOptions.call(this, options)
-    } else if (!options.config && sourceMap && length > 1) {
-      return parseOptions.call(this, options)
-    } else if (options.config && sourceMap && length > 2) {
-      return parseOptions.call(this, options)
+    // Read options from options.config only when options.config is object
+    if (typeof options.config === 'object' &&
+      options.config !== null &&
+      Object.prototype.toString.call(options.config) === '[object Object]'
```
evilebottnawi: Special for `config: true`

lib/index.js
```diff
@@ -47,53 +47,55 @@ module.exports = function loader (css, map) {
...
     }
 
     return postcssrc(rc.ctx, rc.path, { argv: false })
   }).then((config) => {
     if (!config) config = {}
+    if (!config.options) config.options = {}
 
     if (config.file) this.addDependency(config.file)
 
+    let sourceMap = config.options.map || options.sourceMap
```
evilebottnawi: When we load postcss config, she can return `map: undefined`, but in loader we have `sourceMap: true`, because first look on `config.options.map` then on `options.sourceMap`.

lib/options.js
```diff
@@ -11,12 +11,11 @@ module.exports = function parseOptions (params) {
...
+    parser: params.parser,
+    syntax: params.syntax,
+    stringifier: params.stringifier,
+    map: params.sourceMap
```
evilebottnawi: `postcss-load-config` return `map`, we should also return `map` from `parseOptions` (consistently and don't misleading for future options).

lib/index.js
```diff
@@ -47,53 +47,60 @@ module.exports = function loader (css, map) {
...
+        rc.ctx.options = options.config.ctx
+      }
     }
-  }
 
-  const sourceMap = options.sourceMap
+    return postcssrc(rc.ctx, rc.path, { argv: false })
+      .then((config) => {
+        if (!config) config = {}
+        if (!config.options) config.options = {}
```
evilebottnawi: Only `postcss-config-loader` can return undefined

lib/index.js
```diff
@@ -47,53 +47,57 @@ module.exports = function loader (css, map) {
...
 
     return postcssrc(rc.ctx, rc.path, { argv: false })
-  }).then((config) => {
-    if (!config) config = {}
+      .then((config) => {
+        if (!config.options.map) config.options.map = options.sourceMap
 
+        return config
+      })
+  }).then((config) => {
```
evilebottnawi: `config` is always object `parseOptions` return object and `postcss-load-config` also return object

michael-ciniawsky:
I'm not in favour of changing the config loader behaviour to `{config: true }`. This neglects the reason for `postcss.config.js` which is simply to have **no** in `webpack.config.js` for most cases at all. I'm also concerned about merging options from (`webpack.config.js`) && (`postcss.config.js`). It should be a single source of truth and therefore an either or, I don't see the usecase for this, but if besides an edge case or personal perference one exists we can change it.

 >Always load config if your pass config.path, now it is not working as expected if i pass config.path and sourceMap: false my config file not loading -> options.config && !sourceMap && length > 1 (i want to load config and set sourceMap use loader options ), also sometimes i want setup some option from options loader.

`{ sourceMap: false }`  is basically unnessecary (default), just omit it, we need to fix the conditional checks for sure agreed upon that, but without breaking the current behaviour

> Some perf (we don't create rc object because it is not necessary if i use only options without options.config)

👍 
 
> Allow to passing options to postcss throw loader.options, now it is possibly only using postcss config.

What kind of options besides `parser || syntax || stringifier && plugins` ? The only two suitable options for map are `external` or `inline` as an annotation comment, which works via `sourceMap: true || 'inline'`, `from, to` are handled by webpack aswell

evilebottnawi:
@michael-ciniawsky i wan't to use `parser` as option in `postcss-loader` and want to load postcss config. Also i wan't to override plugins, but use config for `parser`, `stringifier` and etc. and many other options.

```js
// TODO
    // Refactor
    if (!options.config && !sourceMap && length) {
      return parseOptions.call(this, options)
    } else if (options.config && !sourceMap && length > 1) {
      return parseOptions.call(this, options)
    } else if (!options.config && sourceMap && length > 1) {
      return parseOptions.call(this, options)
    } else if (options.config && sourceMap && length > 2) {
      return parseOptions.call(this, options)
    }
```

Really don't work in many my project, And I had to roll back (use `postcss-loader` v1). We need to make this explicit, if I want to use configurations (use `path` in `config` or use `true`) then use them. 

This changes works on all my project, testing with `create-react-app`, `angular-starter` and some other popular boilerplate. Btw, current algorithm don't work in `create-react-app` (just disable source map and put config). Seriously it just does not work.

If you do not intend to change the logic of work to explicit, then probably I should do fork, because many in many of my project it is not working as expected, also where i use `create-react-app` it too don't work. It is critical for me.

evilebottnawi:
`{ sourceMap: false } is basically unnessecary (default), just omit it, we need to fix the conditional checks for sure agreed upon that, but without breaking the current behaviour`, i can't change this in many boilerplate, because it is using for all loader from config.

config.js
```
module.exports = {
    ...
    sourceMap: false
    ...
};
```
And in `webpack.config.js` it is setup to all loaders.

michael-ciniawsky:
The current loading algo is crap for sure, but still the behaviour shouldn't change :)

❌ `{ config: true }` or similiar isn't an option (neglect's zero config approach)

**postcss.config.js**
```js
module.exports = ({ file, options, env }) => ({
   // Maybe support options for `map` (PostCSS),
   // but `true` || `inline` handles that fine for the 99% cases,
   // webpack also sets specific requirements here
    sourceMap: options.sourceMap 
});
```

The reason for not supporting `sourceMap` within  `postcss.config.js` is during the fact that, it should work with different PostCSS middlewares and therefore only supports PostCSS options && plugins (middleware agnostic). All other middleware related options are set, where the particular middleware normally provides an interface to do so. `{ sourceMap: true }` is a webpack (`postcss-loader`) specific option and is better defined in `webpack.config.js`

**webpack.config.js**
```js
{
   test: '..',
   use: [
     { loader: 'style-loader', options: { sourceMap: env === 'development' ? true : false} },
     { loader: 'css-loader', options: { sourceMap: env === 'development' ? true : false } },
     // { loader: 'postcss-loader', options: { sourceMap: env === 'development' ? true : false } },
     // Usage worst case scenario
     { 
       loader: 'postcss-loader', 
       options: { config: { ctx: { sourceMap: env === 'development' ? true : false } } } 
     },
     { loader: 'sass-loader', options: { sourceMap: env === 'development' ? true : false} }
   ]
}
```

All we need is to find a better way to filter the loader options and either parse the options, if we have additional ones (`parser &&|| plugins`), or try to load a config file

evilebottnawi:
@michael-ciniawsky Today I will create a table with all the options and behavior that should, so it will be easier to come to a final decision.

evilebottnawi:
@TrySound Maybe you can help us and with this?

evilebottnawi:
@michael-ciniawsky now using `css` inside `sass/less/etc` throw error, https://github.com/insin/nwb/issues/311#issuecomment-304957316 (https://webpack.slack.com/archives/C3NUYSYF4/p1496168402383007)

evilebottnawi:
@michael-ciniawsky also adding any options to `options` (`test: true`) break all loading logic.

michael-ciniawsky:
@evilebottnawi What is the issue in insin/nwb#311 (comment) regarding `postcss-loader`?

> also adding any options to options (test: true) break all loading logic.

Any additional properties should ideally throw `OptionsValidationError`, what is the propose of `{ test: true }` or is the a stub example?

evilebottnawi:
@michael-ciniawsky now it is not throw `OptionsValidationError` :disappointed:  and break loading logic. But is simple
```sass
@import "~swiper/dist/css/swiper.css";
```
in `style-loader` -> `css-loader` -> `postcss-loader` -> `sass-loader`

michael-ciniawsky:
I don't get the issue with the example ? Yeah.. `additionalProperties` is set to true, bc `plugins` prop `{Function}` isn't testeable with `schema-utils` atm. Still additional properties should fail, but with `ValidationError`

evilebottnawi:
@michael-ciniawsky 
```js
webpackConfig.module.rules.push({
        test: /\.s[ac]ss$/,
        use: [
            {
                loader: 'style-loader',
                options: {
                    sourceMap: true
                }
            },
            {
            loader: 'css-loader',
            options: {
                import: true,
                sourceMap: true,
                importLoaders: 2
            }
        },
        {
            loader: 'postcss-loader',
            options: {
                sourceMap: true,
                plugins: () => [require('autoprefixer')]
            }
        },
        {
            loader: 'sass-loader',
            options: { 
                sourceMap: true
            }
        }
    });
```

michael-ciniawsky:
And were does it exactly choke, seems unrelated to `postcss-loader`

evilebottnawi:
@michael-ciniawsky `Module build failed: Error: No PostCSS Config found in:` because tries to load the configuration and `css-loader` (https://github.com/webpack-contrib/css-loader/blob/master/lib/getImportPrefix.js#L12) return only `{ sourceMap: true }`, seems not respect function.

Problems:
1. We need disallow to using `plugins` as function or try to solve respect function option in https://github.com/webpack-contrib/css-loader/blob/master/lib/getImportPrefix.js#L12
2. Throw `OptionsValidationError` on additional argument in options (not a few people are setting the wrong options in an unnecessary place).
3. Also some strange behavior: using `importLoaders` for handling `@import` i got options in `postcss-loader` from `style-loader` -> `css->loader` -> `postcss-loader` (**actual options**) -> `sass-loader` (`test: /\.s[ac]ss$/,`), but i have difference plugins for `sass` and pure `css` (maybe expected behavior), but i was surprised, i think it is using `style-loader` -> `css->loader` -> `postcss-loader` (**expected options**) (`test: /\.css$/`)

evilebottnawi:
@michael-ciniawsky why we use `parseOptions` then:
1. `options.config && !sourceMap && length > 1`
2. `options.config && sourceMap && length > 2`

Can i see example configuration for this use case?

evilebottnawi:
@ai Maybe you can help us solve this problem :disappointed:

michael-ciniawsky:
The `ident` seems to be the problem again

```diff
ExtractTextPlugin.extract({
            fallback: 'style-loader',
            use: [
              {
                loader: 'css-loader',
                options: { importLoaders: 2, sourceMap: true }
              },
              {
                loader: 'postcss-loader',
                options: {
                  config: { ctx: { test: 1 } },
   -               ident: 'postcss',
                  plugins: [ require('postcss-nested')() ],
                  sourceMap: true
                }
              },
              {
                loader: 'sass-loader', options: { sourceMap: true }
              }
            ]
})
```

```
    ERROR in ../~/css-loader?{"importLoaders":2,"sourceMap":true}!../~/postcss-loader/lib?{"config":{"ctx":{"test":1}},"plugins":[null],"sourceMap":true}!../~/sass-loader/lib/loader.js?{"sourceMap":true}!./components/component/index.css
    Module build failed: TypeError: Cannot read property 'postcss' of null
        at Processor.normalize (/Users/Cini/Github/Webpack/__test__/node_modules/postcss-loader/node_modules/postcss/lib/processor.js:136:12)
        at new Processor (/Users/Cini/Github/Webpack/__test__/node_modules/postcss-loader/node_modules/postcss/lib/processor.js:51:25)
        at postcss (/Users/Cini/Github/Webpack/__test__/node_modules/postcss-loader/node_modules/postcss/lib/postcss.js:73:10)
        at Promise.resolve.then.then (/Users/Cini/Github/Webpack/__test__/node_modules/postcss-loader/lib/index.js:137:12)
        at <anonymous>
     @ ../~/css-loader?{"importLoaders":2,"sourceMap":true}!../~/postcss-loader/lib?{"config":{"ctx":{"test":1}},"plugins":[null],"sourceMap":true}!../~/sass-loader/lib/loader.js?{"sourceMap":true}!./App.scss 3:10-313
```

```diff
ExtractTextPlugin.extract({
            fallback: 'style-loader',
            use: [
              {
                loader: 'css-loader',
                options: { importLoaders: 2, sourceMap: true }
              },
              {
                loader: 'postcss-loader',
                options: {
                  config: { ctx: { test: 1 } },
 +               ident: 'postcss',
                  plugins: [ require('postcss-nested')() ],
                  sourceMap: true
                }
              },
              {
                loader: 'sass-loader', options: { sourceMap: true }
              }
            ]
})
```

```
Hash: 730caef604ad1e57d3b2
Version: webpack 2.6.1
Time: 4578ms
        Asset       Size  Chunks             Chunk Names
     index.js  683 bytes   index  [emitted]  index
    vendor.js    5.98 kB  vendor  [emitted]  vendor
    index.css  324 bytes   index  [emitted]  index
 index.js.map  938 bytes   index  [emitted]  index
index.css.map  846 bytes   index  [emitted]  index
vendor.js.map    6.07 kB  vendor  [emitted]  vendor
[./App.js] ./App.js 97 bytes {index} [built]
[./App.scss] ./App.scss 41 bytes {index} [built]
Child extract-text-webpack-plugin:
    [../node_modules/css-loader/index.js?{"importLoaders":2,"sourceMap":true}!../node_modules/postcss-loader/lib/index.js?{"config":{"ctx":{"test":1}},"ident":"postcss","plugins":[null],"sourceMap":true}!../node_modules/sass-loader/lib/loader.js?{"sourceMap":true}!./App.scss] ../~/css-loader?{"importLoaders":2,"sourceMap":true}!../~/postcss-loader/lib?{"config":{"ctx":{"test":1}},"ident":"postcss","plugins":[null],"sourceMap":true}!../~/sass-loader/lib/loader.js?{"sourceMap":true}!./App.scss 1.36 kB {0} [built]
    [../node_modules/css-loader/lib/css-base.js] ../~/css-loader/lib/css-base.js 2.26 kB {0} [built]
        + 1 hidden modules
```

michael-ciniawsky:
The same goes for the `importLoaders` issue, it works with `postcss.config.js`, but chokes on `options.plugins` atm, needs more investigation

evilebottnawi:
@michael-ciniawsky ok then current implementation is correct, only need throw

michael-ciniawsky:
@evilebottnawi Are you intending to update this PR or can I close and open a new one ? 😛

evilebottnawi:
@michael-ciniawsky What is wrong with this PR now? Check on only `plugins` and `config` is incorrectly, i can `use` only `parser` and `stringifier` for parsing and output in other style.

michael-ciniawsky:
`sourceMap`  should be the only way to steup source maps
 `postcss.config.js` is still not **implicitly** loaded

evilebottnawi:
@michael-ciniawsky i.e. we ignore source map from `postcss.config.js`?
It would be nice to see the options under which the config should not boot

michael-ciniawsky:
The sourcemap is mainly because of the loader chain 

```js
[
   { loader: 'style-loader', options: { sourceMap: true } },
   { loader: 'css-loader', options: { sourceMap: true } },
   'postcss-loader' ?sourceMap
   { loader: 'style-loader', options: { sourceMap: true } }
]
```

Also webpack user will then need to know, **how** to config PostCSS Sourcemap Options  instead of `true || 'inline'`

evilebottnawi:
@michael-ciniawsky we already have `inline` or `true` source map. We ignore the source maps from `postcss.config.js`?

michael-ciniawsky:
> we already have inline or true source map. We ignore the source maps from postcss.config.js?

☝️ https://github.com/postcss/postcss-loader/pull/234#issuecomment-308409814

The missing `options.exec` check and #250 + maybe throw if `options.plugins && options.config` are set at the same time, is all we can do in the `v2.0.0` line, but better with a fresh PR :)

evilebottnawi:
@michael-ciniawsky why your close? I will now have to do most of it again, having spent this time, thanks for respecting my time

evilebottnawi:
Main of problems, we have ugly algo and few test, I just do not understand when we need to load config when not, and no specific comments, I'm glad to help and close this problem, but I just do not know what's wrong now :disappointed:

evilebottnawi:
@michael-ciniawsky 
1. Please describe all options when we load config when not or add tests, now behavior is misleading and no one understands except you.
2. What wrong with source map now? We load source map from options, if your have config and `map` we get source map from config, if your have config but don't have `map` we get they from loader `options`?

michael-ciniawsky:
kk mom I'm trying to sum it up

evilebottnawi:
@michael-ciniawsky Please understand me, we have an internal boilerplate config where we can't remove `sourceMap` from source code, it is global variable for all loaders and we need support `postcss.config.js`, it is bug really critical (seems other people have same problem), we can't fix this along time, let's do it **asap**

evilebottnawi:
Let's do it:
1. If we have `exec` and/or `parser` and/or `syntax` and/or `stringifier` and/or `plugins` - we don't load config.
2. If we have `config` - we load config
3. If we don't have options - we load config.
4. sourceMap option does not affect logic, we can use `sourceMap` in 1 and 2 and 3 point.

evilebottnawi:
@TrySound can your review my commit? All tests fine

michael-ciniawsky:
@evilebottnawi Please reduce the PR here only to your proposed `sourceMap` changes

evilebottnawi:
@michael-ciniawsky this PR contain only relevant changed

lib/index.js
```diff
@@ -47,53 +47,55 @@ module.exports = function loader (css, map) {
...
-      },
-      options: {}
-    }
-  }
-
-  if (options.config) {
-    if (options.config.path) {
-      rc.path = path.resolve(options.config.path)
+  Promise.resolve().then(() => {
+    if (options.exec || options.parser || options.syntax || options.stringifier || options.plugins) {
```
michael-ciniawsky: rm

lib/index.js
```diff
@@ -47,53 +47,55 @@ module.exports = function loader (css, map) {
...
+      return parseOptions.call(this, options)
     }
 
-    if (options.config.ctx) {
-      rc.ctx.options = options.config.ctx
+    const rc = {
```
michael-ciniawsky: rm

package.json
```diff
@@ -32,6 +32,7 @@
     "docs": "jsdoc2md lib/index.js > LOADER.md",
     "pretest": "npm run lint && npm run test:build",
     "test": "jest",
+    "test-only": "npm run test:build && jest && npm run clean",
```
michael-ciniawsky: rm

lib/index.js
```diff
@@ -47,53 +47,55 @@ module.exports = function loader (css, map) {
...
-    let options = Object.assign({
+
+    // Disable override `to` option
+    if (config.options.to) delete config.options.to
+    // Disable override `from` option
+    if (config.options.from) delete config.options.from
+    // Disable override `map` option
+    if (config.options.map) delete config.options.map
+
+    let postcssOption = Object.assign({
```
michael-ciniawsky: Why was this renamed, should be `options`

lib/index.js
```diff
@@ -47,53 +47,55 @@ module.exports = function loader (css, map) {
...
-      return parseOptions.call(this, options)
+      if (options.config.ctx) {
+        rc.ctx.options = options.config.ctx
+      }
     }
 
     return postcssrc(rc.ctx, rc.path, { argv: false })
-  }).then((config) => {
-    if (!config) config = {}
+      .then((config) => {
```
michael-ciniawsky: wrong `ident`

evilebottnawi:
@michael-ciniawsky `true` || `inline`

lib/index.js
evilebottnawi: @michael-ciniawsky because `options` related to `loader` (`const options = loaderUtils.getOptions(this) || {}`), `postcssOptions` related to `postcss` options

package.json
evilebottnawi: @michael-ciniawsky How can I run only tests?

lib/index.js
evilebottnawi: @michael-ciniawsky don't understand, please provide example

lib/index.js
evilebottnawi: @michael-ciniawsky we don't need rc if we don't load config

lib/index.js
evilebottnawi: @michael-ciniawsky why? your ugly algo is don't work as expected

michael-ciniawsky:
But that's not the only two options for `map`, what happens of e.g `{ inline: true, annotation: false }` is directly passed via `postcss.config` does it work ?

evilebottnawi:
@michael-ciniawsky In your code https://github.com/postcss/postcss-loader/blob/master/lib/index.js#L99 `true` and `inline` only work. Now algo is same.

lib/index.js
michael-ciniawsky: #259

lib/index.js
michael-ciniawsky: #259

lib/index.js
michael-ciniawsky: the code indentation is off :), at least looks it

evilebottnawi:
@michael-ciniawsky Do you have specific examples of why this code is not working (example of config or failed tests), all tests are passed, or is this a subjective non-constructive mono solution?

lib/index.js
michael-ciniawsky: It works atm due to lexical scope, as long as there is no need to rename it please leave it as is

lib/index.js
evilebottnawi: @michael-ciniawsky 
```js
return postcssrc(rc.ctx, rc.path, { argv: false })
      .then((config) => {
        // Get `sourceMap` option from loader options when no `map` option in `postcss.config.js`
        if (!config.options.map) config.options.map = options.sourceMap

        return config
      })
```
Where?

michael-ciniawsky:
> Do you have specific examples of why this code is not working (example of config or failed tests), all tests are passed, or is this a subjective non-constructive mono solution?

You mean the config resolving here ?

evilebottnawi:
@michael-ciniawsky i mead where algo don't work as expected

evilebottnawi:
@michael-ciniawsky https://github.com/postcss/postcss-loader/pull/259 Thanks for respect my work and my time.

michael-ciniawsky:
Why are you closing this now? 🙃

evilebottnawi:
@michael-ciniawsky because I'm tired of non constructive argument and ignoring, for me it will be faster to create fork. 

Because your say:
> The only two suitable options for map are external or inline as an annotation comment, which works via sourceMap: true || 'inline', from, to are handled by webpack aswell

michael-ciniawsky:
What do you mean by non-constructive argument and ignoring ? Ignoring is just plain wrong... `Disagreement !== Ignoring`

Constrcutive or not is subjective, but

- The PR at first changed the config loading **enterily** and I explained more then once, why I don't think it is a good idea to do it that way 

- For `sourceMap` my concern is that there will be confusion about it, when `postcss.config.js` supports it, but I'm not against it in general (there would still edge case in the PR **as is** )

- The other PR was not meant offensive in any way and I have had no problem by just closing it, I don't care about these things


```js
let options = Object.assign({
      to: file,
      from: file,
      map: sourceMap // <=
        ? sourceMap === 'inline'
          ? { inline: true, annotation: false }
          : { inline: false, annotation: false }
        : false
}, config.options)
```

This would require additional changes to work with **all** possible options `map` PostCSS (`postcss.config.js`) supports

evilebottnawi:
@michael-ciniawsky 
1. your just close my PR without explaining what is wrong - it is non-constructive.
2. Your say 
> The only two suitable options for map are external or inline as an annotation comment, which works via sourceMap: true || 'inline', from, to are handled by webpack aswell
Because i do only `true` and `inline` value supporting for sourceMap
3. `Ignoring` - https://github.com/postcss/postcss-loader/pull/234#issuecomment-303085769, https://github.com/postcss/postcss-loader/pull/234#issuecomment-304947846, https://github.com/postcss/postcss-loader/pull/234#issuecomment-305237199, https://github.com/postcss/postcss-loader/pull/234#issuecomment-306444817
4. No Respect for my time and effort, thanks

 Just don't ignore https://github.com/postcss/postcss-loader/pull/259 and merge PR to solve issue

michael-ciniawsky:
kk, can we please in general agree to take it a bit more relaxed and avoid the childish behaviour ?

> your just close my PR without explaining what is wrong - it is non-constructive.

When I closed the PR is still changed the loading behaviour, that was the reason

> Your say

> The only two suitable options for map are external or inline as an annotation comment, which 
> works via sourceMap: true || 'inline', from, to are handled by webpack aswell
> Because I do only true and inline value supporting for sourceMap


Yes, that's the two ways to handle sourcemaps, `postcss.config.js` supports `{Object}` e.g `{ inline: true, annotation: false }` which `postcss-loader` needs handle aswell then, which is ok if needed

> Ignoring - #234 (comment), #234 (comment), #234 (comment), #234 (comment)

I answer all of them and the second isn't even addressed at me 🙃 

> No Respect for my time and effort, thanks

I hearing this for the third+ time now, what about other people's time ? This is just trying to personify matters to passive aggressively offend  && push others to do what you want?  As I said I don't mind closing my PR's, waste my time in your speaking and in general the least thing I intend to do is to actively disrespect others. Which is disagree I did in anyway, but if you felt that way I apologize

evilebottnawi:
@michael-ciniawsky thanks for apologize, let's first fix unexpected behavior with sourceMap https://github.com/postcss/postcss-loader/pull/259

evilebottnawi:
@michael-ciniawsky and then solve problems: should we respect map from `postcss.config.js` and how we should do this, or now all ok

michael-ciniawsky:
Yes, but we need to add a check / refactor https://github.com/postcss/postcss-loader/blob/master/lib/index.js#L99 properly to support 'full' PostCSS Sourcemap Options `{Object}`, also the `config.options.from/comfig.options.to`  sanitizing you added was a good idea

michael-ciniawsky:
Also check if some possible PostCSS Sourcemap Options doesn't play well with webpack (I don't know 💯 yet)

PR #245, opened 2017-05-30
feat(index): add ICSS support (`url()` && `@import`)

### LINKED ISSUE #228: [2.1.0] Add `url()` handling (`postcss-icss-url`)
### `postcss-icss-url`

https://github.com/TrySound/postcss-icss-url

### `postcss-loader`

```
// TODO
```

michael-ciniawsky:
Fixes #228 

### Feature

- [x] add [`postcss-icss-url`](https://github.com/css-modules/postcss-icss-url) for `url()` handling
- [x] add [`postcss-icss-import`](https://github.com/css-modules/postcss-icss-import) for `@import` handling

#### `url()`

```css
.url {
   background-url: ('./img.png');
}
```

```css
:import('./img.png'){
  __url_0: default
}

.url {
  background-url: url(__url_0)
}
```

#### `@import`

```css
@import './style.css';
```

```css
:import('./style.css') {
  import: default
}
```

### TODO

- [ ] update README
- [ ] update tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,8 @@
   "dependencies": {
     "loader-utils": "^1.x",
     "postcss": "^6.x",
+    "postcss-icss-import": "^0.x",
+    "postcss-icss-url": "^0.x",
     "postcss-load-config": "^1.x",
     "schema-utils": "^0.x"
   },
@@ -23,7 +25,7 @@
     "jsdoc-to-markdown": "^3.x",
     "postcss-js": "^0.x",
     "standard": "^10.x",
-    "standard-version": "^4.0.0",
+    "standard-version": "^4.x",
     "sugarss": "^1.x",
     "webpack": "^2.x"
   },
--- a/lib/index.js
+++ b/lib/index.js
@@ -93,6 +93,14 @@ module.exports = function loader (css, map) {
     if (config.file) this.addDependency(config.file)
 
     let plugins = config.plugins || []
+
+    // Add ICSS Support
+    // url() && @import
+    plugins.concat([
+      require('postcss-icss-url')(),
+      require('postcss-icss-import')()
+    ])
+
     let options = Object.assign({
       to: file,
       from: file,


TrySound:
I'd like to make them disablable (fogot the word)

michael-ciniawsky:
@TrySound Via extra `loader` option ? How strong is the perf impact approximately ? Does it potentially clash with other plugins ? One option for implicit config could be 

```js
...
if (this.loaderIndex !== 0) plugins.concat([...icss]) // `css-loader` in loader chain ? push : skip
...
```

michael-ciniawsky:
We need to think about config for CSS Modules to in near future aswell, best would be to have plugin config only and `postcss-loader` handles the rest implicitly (`fetch`, `resolve`, etc)

package.json
```diff
@@ -23,7 +25,7 @@
     "jsdoc-to-markdown": "^3.x",
     "postcss-js": "^0.x",
     "standard": "^10.x",
-    "standard-version": "^4.0.0",
+    "standard-version": "^4.x",
     "sugarss": "^1.x",
     "webpack": "^2.x"
```
Kovensky: These version numbers look wrong. You're already using the semver caret; using the `x` placeholder probably only works by accident.

TrySound:
@michael-ciniawsky Not everybody want to use this url processor and the same with imports.

evilebottnawi:
@michael-ciniawsky i agree with @TrySound, not all want to using these plugins, best way for this add docs about this.

TrySound:
Maybe we better need configurable plugin pack for css-modules.

evilebottnawi:
@TrySound Sounds good too, but we definitely should not use them always in `postcss-loader`

TrySound:
Yeah. I'm about it.

michael-ciniawsky:
Yeah I'm all in not adding them if avoidable, but I'm seriously worried about the downgrade between old && new CSS pipeline in terms of `url() && @import`handling then (more config). This shouldn't be a concern for average users not familiar with any webpack 'internals' at all. Without these plugins none of the currently working _runner specific_ stuff works anymore (`url-loader/file-loader` [`url()`]) && `@import => import css from './import.css` then. CSS Modules as plugin explicitly setup 👍, but if `url() && @import` don't work `out-of-the-box` one could use the `raw-loader` aswell 😛. This isn't _really_ and opt-in. Plugin Order is also a potential issue then.

- Why is explicit config better here ?
- What are the usecases where ICSS isn't needed and are the common ?

ICSS is bsically a **requirement** not an opt-in and in the grey area between 'normal' PostCSS Config and webpack specific, with a high tendency to be webpack specific => PostCSS Runner (IMHO)

package.json
michael-ciniawsky: Does the exact `minor.patch` really matter when the `^` is used ?

TrySound:
Plugin order won't be a problem. Each stuff should have only own responsibility. I for example always disable url in css loader. Maybe just because it has some problems, but still a lot of people converts their build steps and may have different url process.

michael-ciniawsky:
```js
const plugins = [
  require('postcss-icss-url')
  require('postcss-nested')
]
```

Would this still work for e.g ?

Yeah it would need an additional option at least for `url()`, in case there should be no rewrite at all 😕. So better remove `postcss-icss-url` agreed. Hmm... this all so unconvenient... 😛

TrySound:
Well I meant icss plugins order. But yeah this case will work.

TrySound:
I think plugin pack is good solution about order. Just pass pack in the end.

michael-ciniawsky:
kk I write initial docs and open new PR docs only for now

clydin:
Is there a reason ICSS is a requirement?  Making a non-standard (albeit quite novel) extension to CSS a requirement doesn't seem like a good long-term plan.  There's bound to be compatibility issues in the future.  For the CSS Module use cases (which do look quite nice), ICSS should be a supported option but there should be a path that doesn't require its use.  Further, the ICSS concept still relies on the _special string marker and replace_ strategy.  (Some form of this is inevitable if the goal is to support url/file-loader but it could be cleaner).  Especially since the ICSS postcss plugins are essentially adding additional text content to the stylesheets to have them almost immediately removed.

Was a side channel path considered as an alternative or companion?  Postcss plugins could analyze any `url()` or `@import` and then generate postcss messages containing the asset and the actual text to match and replace (potentially with position hint information to optimize replacement) and eliminate the special string marker _hack_ (although that could still be used).  The `postcss-loader` could than add the collected data to the generated webpack module for the stylesheet.  Downside is this would currently break the `raw-loader`-like functionality; although that functionality is not overly useful in this scenario in general (maybe an option for more basic scenarios?).   Also with the pending move of stylesheets becoming first-class, this could potentially be remedied (allow passing ancillary data through the callback?).  Upside is that it removes the back and forth CSS manipulation and gives webpack actual structured data to use directly; which should provide a more direct stylesheet processing pipeline and a decent performance improvement.  Also it appears that postcss plugins already exist to extract the ICSS import/export pseudo-selector information, so they could be used if the ICSS features were desired and eliminate the need for webpack to support them directly.

evilebottnawi:
@clydin your can pass icss `url` and `import` plugins to end and avoid problems with other plugins. Also your can create own `icss-loader` and insert them after `css-loader`.

clydin:
Fundamentally, my concern is that webpack directly using ICSS seems unnecessary, inefficient (although less inefficient than the current css-loader), and non-standard.  On the javascript front, proprietary methods are being relegated as legacy (e.g., `require.ensure`).  However, stylesheets appear to be going in the opposite direction.

evilebottnawi:
@clydin ICSS is new gulp of fresh css. Now css don't have module system, no import and export, but js have support modules. This is probably one of main reason why ICSS was created. Don't worry about `inefficient`, it is faster and easy. You continue to write on CSS (with own postcss plugins and etc). ICSS v1 already used in `css-loader`. New css-loader have only support ICSS v2 and it is great, fast and easy to use. Wait and enjoy ICSS, do not worry.

clydin:
Thank you for the response.  However, with the widespread use of webpack (and postcss-loader) and further use in major downstream integration projects, I was hoping for a more technical discussion on the merits of the design decision based on both common and advanced use cases. As well as some insight into the analysis of alternatives aspects of the re-architecting of a major webpack subsystem.

Although this PR is probably not the appropriate place for such a conversation and I apologize for the inappropriate segue.

evilebottnawi:
@clydin your can join to slack general channel https://webpack.slack.com/messages/C1LUX2DS9/ and ask sokra about related questions.

Issue #250, opened 2017-06-06
[2.0.5] Explicit `false` not working (`options.sourceMap`)


leebenson:
I have a function that dynamically sets whether sourceMaps should be included (amongst other options):

```js
getExtractCSSLoaders(extractCSS, sourceMap = false) {
  return (function* loadCss() {
    for (const loader of css.loaders) {
      // Iterate over CSS/SASS/LESS and yield local and global mod configs
      for (const mod of css.getModuleRegExp(loader.ext)) {
        yield {
          test: new RegExp(mod[0]),
          loader: extractCSS.extract({
            use: [
              {
                loader: 'css-loader',
                query: Object.assign({}, css.loaderDefaults, {
                  sourceMap,
                }, mod[1]),
              },
              // 'postcss-loader',
              {
                loader: 'postcss-loader',
                options: {
                  sourceMap: false,
                },
              },
              ...loader.use,
            ],
            fallback: 'style-loader',
          }),
        };
      }
    }
  }());
},
```

Explicitly setting `sourceMap: false` seems to kill/bypass postcss:

```js
// Explicitly setting `sourceMap` to the default `false` kills it...
{
  loader: 'postcss-loader',
  options: {
    sourceMap: false,
  },
},
```

Setting it to true works fine:

```js
{
  loader: 'postcss-loader',
  options: {
    sourceMap: false,
  },
},
```

And ignoring the option is also fine:

```js
{
  loader: 'postcss-loader',
},
```

A temporary workaround is to do a slightly uglier:

```js
{
  // ...
  sourceMap ? {
    loader: 'postcss-loader',
    options: {
      sourceMap: true,
    },
  } : 'postcss-loader',
  // ...
}
```

michael-ciniawsky:
Yep, it's a know issue with the current config algo using `!sourceMap` internally, I try to fix it as soon as possible, but please omit it for now

michael-ciniawsky:
Fixed && Released in [`2.0.6`](https://github.com/postcss/postcss-loader/releases/tag/v2.0.6) please test and verify :)

leebenson:
Works great, thanks @michael-ciniawsky !

PR #259, opened 2017-06-14
fix(index): simplify config loading behaviour

michael-ciniawsky:


evilebottnawi:
@michael-ciniawsky I will not argue again because it's useless, just merge this pr to solve https://github.com/postcss/postcss-loader/issues/250

michael-ciniawsky:
Fixes #250

evilebottnawi:
@michael-ciniawsky let's do release :+1: after merge

PR #269, opened 2017-06-29
fix(index): runaway promise

mcjfunk:
I'm getting a runaway promise warning:
```
Warning: a promise was created in a handler at /Users/jmclean/projects/node-graphql-demo/node_modules/postcss-loader/lib/index.js:173:16 but was not returned from it, see http://goo.gl/rRqMUw
    at new Promise (/Users/jmclean/projects/node-graphql-demo/node_modules/bluebird/js/release/promise.js:77:14)
```
Seems that the loader returns undefined in certain situations, falling back to null fixes it and seems to be the recommended practice from the Bluebird docs: http://goo.gl/rRqMUw

There was a similar fix, about a year ago. #95

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -165,7 +165,7 @@ module.exports = function loader (css, map) {
            * @param {String} result Result (JS Module)
            * @param {Object} map    Source Map
            */
-          return cb(null, `module.exports = ${JSON.stringify(css)}`, map)
+          return cb(null, `module.exports = ${JSON.stringify(css)}`, map) || null
         }
         /**
          * @memberof loader
@@ -175,7 +175,7 @@ module.exports = function loader (css, map) {
          * @param {String} css  Result (Raw Module)
          * @param {Object} map  Source Map
          */
-        return cb(null, css, map)
+        return cb(null, css, map) || null
       })
   }).catch((err) => {
     return err.name === 'CssSyntaxError' ? cb(new SyntaxError(err)) : cb(err)


michael-ciniawsky:
@mcjfunk I which usecases does this occur? It might not throw, but looks odd tbh :). e.g the current file would be lost then right?

mcjfunk:
Using the configuration below running the `webpack-dev-middleware` with a hapi.js server.
```
{
    test: /\.less$/,
    use: [
        {loader: 'css-loader', options: {importLoaders: 1}},
        {loader: 'postcss-loader', options: {plugins: [autoprefixer({browsers}), cssnano({safe: true})]}},
        'less-loader'
    ]
},
```
There is no promise warning outside of the 'webpack-dev-middleware` workflow.

With (and without) the fix, the styles are correctly processed and injected onto the page. The fix quiets down the promise warning that happens on every less file it processes.

Looks like others have had to implement a similar fix:
https://github.com/vuejs/vue-loader/issues/625

Thanks!

evilebottnawi:
@mcjfunk seems related to `less-loader`, can your create minimum reproducible test repo?

mcjfunk:
I won't have a chance to put together a test repo anytime soon :(.

I tried a test with a CSS file instead of less, and removed the less-loader. The warnings were still present.

When I remove the `css-loader`, the warnings go away.

michael-ciniawsky:
I need a reproducible case for this to land it in `postcss-loader` directly, as long as it isn't coming from the loader itself, I would like to avoid adding workarounds. `css-loader` will get an overhaul soon. It is dog slow and buggy atm. If you tracked it to `css-loader` open an issue there aswell please either @evilebottnawi or I can fix it there also

lib/index.js
```diff
@@ -165,7 +165,7 @@ module.exports = function loader (css, map) {
            * @param {String} result Result (JS Module)
            * @param {Object} map    Source Map
            */
-          return cb(null, `module.exports = ${JSON.stringify(css)}`, map)
+          return cb(null, `module.exports = ${JSON.stringify(css)}`, map) || null
```
michael-ciniawsky:
```js
cb(null, `module.exports = ${JSON.stringify(css)}`, map)
return null
```

lib/index.js
```diff
@@ -175,7 +175,7 @@ module.exports = function loader (css, map) {
          * @param {String} css  Result (Raw Module)
          * @param {Object} map  Source Map
          */
-        return cb(null, css, map)
+        return cb(null, css, map) || null
```
michael-ciniawsky:
```js
cb(null, css, map)
return null
```

DmitriWolf: I'm working on a React app at Homeaway that spits out a long string of broken promise warnings every time I build, and either of these two fixes makes them all go away.
Seems like a trivial fix to a constantly annoying problem. 
Please merge one or the other. Thank you.

michael-ciniawsky: @DmitriWolf Are you also using `bluebird` ?

DmitriWolf: We are!

timmolendijk:
Returning `null` there instead of `undefined` is definitely the way to go!

Previous versions (1.3.3 f.e.) returned `null` there, so let's stick to that behavior.

michael-ciniawsky:
Released in [v2.0.7](https://github.com/postcss/postcss-loader/releases/tag/v2.0.7)

PR #314, opened 2017-11-24
test: ident && config not working

solcik:
Hello,

I was just trying to figure out for a long time now, why my postcss-config is not taken into account, when using postcss-loader with webpack.
I dug into the loader and found this: https://github.com/postcss/postcss-loader/blob/master/lib/index.js#L53-L65, that is why my setup is not working.
Variable length has in this case value of 1, therefore return at line L65 and not using the postcssrc.

```
{
  loader: 'postcss-loader',
  options: {
    sourceMap: true,
    ident: 'postcss',
    config: {
      path: paths.DIR.ROOT,
    },
  },
};
```

### DIFF
--- a/test/options/__snapshots__/config.test.js.snap
+++ b/test/options/__snapshots__/config.test.js.snap
@@ -2,6 +2,8 @@
 
 exports[`Options Config - {Object} 1`] = `"module.exports = \\"a { color: rgba(255, 0, 0, 1.0) }\\\\n\\""`;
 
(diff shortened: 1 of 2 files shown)


michael-ciniawsky:
Thx 👍 Fixed in #315, you only need the `ident` if you are using a `{Function} || require()` within`options` (normally when using `options.plugins` instead of a `postcss.config.js`)

michael-ciniawsky:
Fixed && Released in  [`v2.0.9`](https://github.com/postcss/postcss-loader/releases/tag/v2.0.9) 🎉

PR #326, opened 2018-01-03
fix(index): copy loader `options` before modifying

clydin:
As per the `loader-utils` documentation ([here](https://github.com/webpack/loader-utils#getoptions)),  the options object should not be modified directly.  Doing so causes the loader parameter of a `plugins` option function to only be run once at the first use of the loader.  This effectively causes the loader parameter to be useless as it cannot be used to retrieve information about the actual resource being used (i.e., `loader.resourcePath`, etc.) instead it will always point to the first resource the loader handled.  This PR makes a copy of the options object returned from `getOptions`.

### `Type`
---

- [X] Fix

### `SemVer`
---

- [X] Bug (:label: Patch)

### `Issues`
---

### `Checklist`
---

- [X] Lint and unit tests pass with my changes

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -40,7 +40,7 @@ const SyntaxError = require('./Error')
  * @return {cb} cb      Result
  */
 module.exports = function loader (css, map) {
-  const options = loaderUtils.getOptions(this) || {}
+  const options = Object.assign({}, loaderUtils.getOptions(this))
 
   validateOptions(require('./options.json'), options, 'PostCSS Loader')
 
@@ -51,7 +51,7 @@ module.exports = function loader (css, map) {
 
   Promise.resolve().then(() => {
     const length = Object.keys(options)
-      .filter((option) =>  {
+      .filter((option) => {
         switch (option) {
           // case 'exec':
           case 'ident':


lib/index.js
```diff
@@ -51,7 +51,7 @@ module.exports = function loader (css, map) {
 
   Promise.resolve().then(() => {
     const length = Object.keys(options)
-      .filter((option) =>  {
+      .filter((option) => {
```
michael-ciniawsky: 😅 👍

michael-ciniawsky:
@clydin Did you test with `postcss.config.js` aswell ? This shouldn't be an issue here I assume...
Can you give an small usage example where this is reproducible ? (When does exactly is this causing undesired behavior/breaks ?)

clydin:
The issue stems from this line which is modifying the options object: https://github.com/postcss/postcss-loader/blob/master/lib/options.js#L5

Since that line changes the `plugins` property value from the function to the returned array and that options object is still used and kept by webpack, the function will never be called again as the `plugins` property is now an array for the rest of the webpack compilation.  This essentially captures the passed in loader context object inside the returned plugins array.  So the example in the readme (https://github.com/postcss/postcss-loader#plugins) for accessing `loader.resourcePath` only works for the first resource loaded.  Every subsequent file processed with the loader will have the same first `loader.resourcePath`.

michael-ciniawsky:
Released in [v2.0.10](https://github.com/postcss/postcss-loader/releases/tag/v2.0.10) Thx

PR #345, opened 2018-03-17
feat(index): merge `webpack` and `rc` options (`options.postcssrc`)

Fnatte:
*Some background:*

I recently upgraded from 1.3.3 to 2.1.1. It broke my build and it took me some time to figure out why. It turns out that I had a bad option `sourceComments` passed to postcss-loader, which probably had been hanging around since a previous switch from sass to postcss. The extra option would cause postcss-loader to skip postcss.config.js entirely because of options length-check in `index.js`:

https://github.com/postcss/postcss-loader/blob/1e0cade20890a150e2187172616ffb523cb17bab/lib/index.js#L67-L69

The options passes validation because additionalProperties is set to true in `options.json`.

*Proposed change*

The problem above could probably be solved by setting additionalProperties to false. But it seems to me that the expected thing to do here is to merge the option sources (both webpack and postcss.config.js), just like babel-loader. This does not only solve the background problem, but is also a handy feature.

*Change*

* Add boolean postcssrc option that control wheather or not to parse and
use postcss.config.js
* Merge options from webpack with the options defined in
postcss.config.js
* Update some tests with postcssrc option set to false when it was
expected
* Add tests for postcssrc option and a test for unknown options

Any thoughts?

### `Type`
---

- [X] Feature
- [X] Refactor

### `SemVer`
---

- [x] Breaking Change (:label: Major)

### `Checklist`
---

- [X] Lint and unit tests pass with my changes

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -5,10 +5,10 @@ const path = require('path')
 const loaderUtils = require('loader-utils')
 
 const parseOptions = require('./options')
+const parseRc = require('./rc')
 const validateOptions = require('schema-utils')
 
 const postcss = require('postcss')
-const postcssrc = require('postcss-load-config')
 
 const SyntaxError = require('./Error')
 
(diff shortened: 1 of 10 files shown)


michael-ciniawsky:
> The problem above could probably be solved by setting additionalProperties to false

Yep, we should definitely do that :)

Fnatte:
> Declaring related options in two different places and adding additional complexity should be avoided whenever possible and is bad practice in general imho.

I am totally with you on that point. My thinking was that it was better to stick with how (most) other loaders work. I may be wrong, and I have not a lot to back this up with except babel-loader, but I believe that many loaders (that also allow rc-files) have support for merging. On the other hand, it's probably a good idea to disallow merging as long as we can prevent misunderstanding.

One of the somewhat weird behaviours (although justified) is that configuration done in webpack options apply to either the actual loader or postcss. Also, it is not obvious to me that adding `syntax: 'sugarss'` to webpack config will cause postcss.config.js not to be parsed at all.

I do not have any good answers to your questions, so maybe that is enough to close this PR. Do you think a warning message is appropriate when an option is causing postcss.config.js to be skipped?

I will open another PR regarding `additionalProperties`.

michael-ciniawsky:
> I am totally with you on that point. My thinking was that it was better to stick with how (most) other loaders work. I may be wrong, and I have not a lot to back this up with except babel-loader, but I believe that many loaders (that also allow rc-files) have support for merging. On the other hand, it's probably a good idea to disallow merging as long as we can prevent misunderstanding.

`babel` has lot of issues with this aswell and their is a third flavor in the making already :) https://github.com/babel/babel/pull/7358. `babel` has the benefit that it's plugins work differently and don't rely that heavily on exact plugin order (In-order AST Reducing (`postcss`) vs. Visitor Pattern (`babel`))

> One of the somewhat weird behaviours (although justified) is that configuration done in webpack options apply to either the actual loader or postcss. Also, it is not obvious to me that adding syntax: 'sugarss' to webpack config will cause postcss.config.js not to be parsed at all.

Things like `ident` && `sourceMap` etc are `webpack` loader specifics which in terms of consistency is a valid point of critique, but that won't be resolvable with filtering these props out before passing it to `postcss` in either way :). E.g declaring `sourceMap` in `webpack.config.js` still works with `postcss.configjs` since it's `webpack` loader specific and has restrictions compared to `postcss` sourcemap options. If this is too unclear/confusing in your opinion it should be better documented in the README instead

> I do not have any good answers to your questions, so maybe that is enough to close this PR. Do you think a warning message is appropriate when an option is causing postcss.config.js to be skipped?

It's edgy in the nuances and still not perfect even with  a 'clean' approach taken by separation :)

> Do you think a warning message is appropriate when an option is causing postcss.config.js to be skipped?

The problem I see with warnings is that `webpack` doesn't offer a way to only display them once, so everytime a `.css` file is processed by the loader the warning will be displayed which can easily stack up to hundreds of warnings... Still agreeing that there should be some hint in this case as one gets and `{Error}` in case both no `webpack.config.js` options or `postcss.config.js` is found 🤔

PR #352, opened 2018-03-28
feat(index): expose `options` as a  `{Function}` (`options.parser`)

### LINKED ISSUE #351: `parser` is ambiguously handled
The documentation lists the valid types for `parser` as `{String|Function}` in the "Syntaxes" section, but it actually only accepts `{String|Object}`. It's mentioned as `{String|Object}` in the section above.

### `Details`
---

Changing `options.json` to allow a function does work, so I'm not sure whether the code or the documentation is wrong here.
Apart from this, there's no description of the non-string types and how they should be written.

The reason I stumbled upon this is that I'm trying to use the dynamic parser functionality that's currently possible when using a postcss.config.js, but doing it using only my webpack.config; for this reason, I'd prefer the code to follow the documentation, instead of the other way around.

Cybolic:
Fixes issue #351, adding support for `parser` to be a function, as mentioned in the existing documentation.

### `Type`
---

- [x] Feature

### `SemVer`
---

- [x] Feature (:label: Minor)

### `Issues`
---

- Fixes #351

### `Checklist`
---

- [x] I have added tests that prove my fix is effective/works
- [x] Any dependent changes are merged and published in downstream modules

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,7 +246,7 @@ You can use a dynamic setup similar to the one listed in the [`ctx` section](#co
   test: /\.[cs]ss$/,
   use: [
     ...,
-    { loader: 'postcss-loader', options: { parser: (css, opts) => opts.from.match(/\.sss$/) ? sugarss.parse(css) : css } }
+    { loader: 'postcss-loader', options: { parser: (css, opts) => opts.from.match(/\.sss$/) ? sugarss.parse(css) : postcss.parse(css) } }
   ]
 }
 ```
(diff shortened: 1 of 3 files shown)


evilebottnawi:
@Cybolic tests failed, can you fix it?

Cybolic:
The test fail on comparing the snapshot, not on the actual code. I'm afraid I'm not familiar enough with the test setup to know what needs to be done to fix the error.

michael-ciniawsky:
Closing due to inactivity for now, feel free to reopen :)

PR #365, opened 2018-06-27
fix(index): emit `warnings` as an instance of `{Error}` instead of `{String}`

### LINKED ISSUE #350: postcss warnings print "(Emitted value instead of an instance of Error)"
### `Details`
---

When using a postcss plugin that may report warnings (e.g. postcss-custom-properties when it can't find the var definition, or stylelint in postcss plugin mode), the warnings are not emitted according to the webpack3+ API that requires that all warnings be instances of Error.

### `Error (Logs|Stacks)`
---

```
WARNING in .../node_modules/css-loader??ref--13-1!.../node_modules/postcss-loader/lib??ref--13-2!./.../style.css
(Emitted value instead of an instance of Error) stylelint: ./.../style.css:35:20: Expected a trailing semicolon (declaration-block-trailing-semicolon)
 @ ./.../style.css 2:14-159 21:1-42:3 21:157-42:2 22:19-164
 @ ./.../index.tsx
...
 @ multi (webpack)-dev-server/client?https://localhost:8080 webpack/hot/dev-server .
```

### `Reproduction [Code]`
---

A stylelint-as-plugin setup which config just `{ "extends": "stylelint-config-standard" }` is probably enough to trigger warnings even on an empty CSS file.

---

To clarify, the problem is that whenever a warning from postcss is printed, it is always prefixed by a warning from webpack saying `(Emitted value instead of an instance of Error)`, although I could also add a feature request for nicely formatted errors like `postcss-reporter` does :)

maxleaver:
Warnings are currently sent to webpack as strings. However, Webpack's emitWarning function expects an error, [as described in their API](https://webpack.js.org/api/loaders/#this-emitwarning). This results in the errors described in #350. This PR simply changes the warnings emitted by postcss-loader from strings to an error instance.

### `Type`
---

- [x] Fix

### `SemVer`
---

- [x] Bug (:label: Patch)

### `Issues`
---

- Fixes #350

### `Checklist`
---

> _Put an `x` in the boxes that apply._

- [ ] Lint and unit tests pass with my changes

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -143,7 +143,7 @@ module.exports = function loader (css, map, meta) {
     return postcss(plugins)
       .process(css, options)
       .then((result) => {
-        result.warnings().forEach((msg) => this.emitWarning(msg.toString()))
+        result.warnings().forEach((msg) => this.emitWarning(new Error(msg.toString())))
 
         result.messages.forEach((msg) => {
           if (msg.type === 'dependency') this.addDependency(msg.file)


lib/index.js
```diff
@@ -143,7 +143,7 @@ module.exports = function loader (css, map, meta) {
     return postcss(plugins)
       .process(css, options)
       .then((result) => {
-        result.warnings().forEach((msg) => this.emitWarning(msg.toString()))
+        result.warnings().forEach((msg) => this.emitWarning(new Error(msg.toString())))
```
michael-ciniawsky: I need to check the Postcss API docs, but I think `warnings` can contain `location` info and a code frame if the plugin author used the appropiated PostCSS API. If that's the case it would be better to reuse the code located in `lib/Error.js` and create a new class `LoaderWarning` (`lib/Warning.js`) based upon that to be consistent in terms of formatting

michael-ciniawsky:
Closing in favor of https://github.com/postcss/postcss-loader/pull/384/commits/6eb82be7e7115f010946c1df1ab2072b59975e7c Thx

PR #371, opened 2018-07-14
docs(README): clarify `path` option description (`options.config.path`)

y-gagar1n:
The description for `path` option in **README.md** is a bit misleading. It says `PostCSS Config Path` while in fact it sets only the directory where **postcss.config.js** will be looked for. 

It leads to a confusion, because users may think that they can set arbitrary name for their config file.

### `Type`
---
- [x] Docs

### `SemVer`
---
- [x] Bug (:label: Patch)

### `Issues`
---

### `Checklist`
---

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ If you use JS styles without the [`postcss-js`][postcss-js] parser, add the `exe
 
 You can manually specify the path to search for your config (`postcss.config.js`) with the `config.path` option. This is needed if you store your config  in a separate e.g `./config || ./.config` folder.
 
-> ⚠️  Otherwise it is **unnecessary** to set this option and is **not** recommended. You **can't** set other name for **postcss.config.js**, this option only allows you to specify the folder where it will be looked for.
+> ⚠️  Note that you **can't** use a **filename** other than the [supported config formats] (e.g `.postcssrc.js`, `postcss.config.js`), this option only allows you to manually specify the **directory** where config lookup should **start** from
 
 **webpack.config.js**
 ```js
@@ -152,11 +152,14 @@ You can manually specify the path to search for your config (`postcss.config.js`
   loader: 'postcss-loader',
   options: {
     config: {
-      path: 'path/to/postcss.config.js'
+      path: 'path/to/.config/' ✅
+      path: 'path/to/.config/css.config.js' ❌
     }
   }
 }
 ```
+
+[supported config formats]: https://github.com/michael-ciniawsky/postcss-load-config#usage
 
 #### `Context (ctx)`
 
--- a/README.md
+++ b/README.md
@@ -137,14 +137,14 @@ If you use JS styles without the [`postcss-js`][postcss-js] parser, add the `exe
 
 |Name|Type|Default|Description|
 |:--:|:--:|:-----:|:----------|
-|[`path`](#path)|`{String}`|`undefined`|PostCSS Config Path|
+|[`path`](#path)|`{String}`|`undefined`|PostCSS Config Directory|
 |[`context`](#context)|`{Object}`|`undefined`|PostCSS Config Context|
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -137,14 +137,14 @@ If you use JS styles without the [`postcss-js`][postcss-js] parser, add the `exe
...
-|[`path`](#path)|`{String}`|`undefined`|PostCSS Config Path|
+|[`path`](#path)|`{String}`|`undefined`|PostCSS Config Directory|
 |[`context`](#context)|`{Object}`|`undefined`|PostCSS Config Context|
 
 #### `Path`
 
 You can manually specify the path to search for your config (`postcss.config.js`) with the `config.path` option. This is needed if you store your config  in a separate e.g `./config || ./.config` folder.
 
-> ⚠️  Otherwise it is **unnecessary** to set this option and is **not** recommended
+> ⚠️  Otherwise it is **unnecessary** to set this option and is **not** recommended. You **can't** set other name for **postcss.config.js**, this option only allows you to specify the folder where it will be looked for.
```
michael-ciniawsky: > ⚠️ Note that you **can't** use other **filename** other then the [supported config formats](https://github.com/michael-ciniawsky/postcss-load-config#usage) (e.g `.postcssrc.js`, `postcss.config.js`), this option only allows you to manually specify the **directory** where config lookup should **start** from

**webpack.config.js**
```js
{
   config: {
      path: 'path/to/.config/' ✅ 
      path: 'path/to/.config/css.config.js' ❌ 
   }
}
```

y-gagar1n:
Fixed

README.md
```diff
@@ -144,20 +144,23 @@ If you use JS styles without the [`postcss-js`][postcss-js] parser, add the `exe
 
 You can manually specify the path to search for your config (`postcss.config.js`) with the `config.path` option. This is needed if you store your config  in a separate e.g `./config || ./.config` folder.
 
-> ⚠️  Otherwise it is **unnecessary** to set this option and is **not** recommended. You **can't** set other name for **postcss.config.js**, this option only allows you to specify the folder where it will be looked for.
+> ⚠️  Note that you **can't** use a **filename** other than the [supported config formats] (e.g `.postcssrc.js`, `postcss.config.js`), this option only allows you to manually specify the **directory** where config lookup should **start** from
```
michael-ciniawsky: Please still leave the previous warning 

> ⚠️  Otherwise it is **unnecessary** to set this option and is **not** recommended.

y-gagar1n:
Fixed

y-gagar1n:
Turns out those ✅ and ❌ symbols are shown as squares on Chrome in Debian.

michael-ciniawsky:
Can you post a screenshot and does this also happen with other symbols (e.g the various ⚠️) ? Chrome only or in other browsers aswell ?

y-gagar1n:
Hmm, must've been some specific version of Chrome or something. Seen that on my work machine today. Now I check on my personal machine (also Chrome and Debian) and everything is OK. Will discover it more tomorrow.

y-gagar1n:
It's caused by the lack of some system fonts on my Debian. I've made a little research and found out that there are more widely supported symbols for 'cross' and 'check' marks. The downside is that they are not so fancy looking.

Here they are with their supported fonts:

Used now:

 ✅ http://www.fileformat.info/info/unicode/char/2705/fontsupport.htm
❌ http://www.fileformat.info/info/unicode/char/274c/fontsupport.htm

Recommended:

✔ https://www.fileformat.info/info/unicode/char/2714/fontsupport.htm
✖ https://www.fileformat.info/info/unicode/char/2716/fontsupport.htm

y-gagar1n:
And here is the screenshot you asked for. Same behavior in Chrome and Opera. 

Firefox works fine.

PR #384, opened 2018-08-04
chore(package): v3.0.0

### LINKED ISSUE #350: postcss warnings print "(Emitted value instead of an instance of Error)"

### LINKED ISSUE #374: Use PostCSS 7
There are no real breaking changes, we just dropped Node.js 4

### LINKED ISSUE #385: [v3.0.0] Check for supported ES2015 Features and update the code base
`postcss-loader >= v3.0.0` drops support for `node >= v4.0.0`, therefore all [ES2015 Features supported by `node >= v6.0.0`](https://node.green) can be used by `postcss-loader >= v3.0.0`

> ℹ️  Development happens on the `develop` branch, please open PR's against this branch if you want to contribute :)

### `Details`

- None

### `Error (Logs|Stacks)`
 
- None

### `Reproduction [Code]`

- None

### `Environment`

 - None

michael-ciniawsky:
#### `Commit Message Summary (CHANGELOG)`

```bash
BREAKING CHANGE: requires `node >= v6.0.0`
```

### Type

- [x] Chore

### Issues

- Fixes #350
- Fixes #374 
- Fixes #385
- Closes #365

### SemVer

- [x] Breaking Change (:label: Major)

### Checklist

- [x] I have added tests that prove my fix is effective/works
- [x] Any dependent changes are merged and published in downstream modules


src/index.js
```diff
@@ -2,15 +2,14 @@
 
 const path = require('path')
 
-const loaderUtils = require('loader-utils')
-
-const parseOptions = require('./options')
-const validateOptions = require('schema-utils')
+const { getOptions } = require('@webpack-utilities/loader')
```
evilebottnawi: Don't use this package, we don't have time support this package, also he use own logic for parse query, in webpack@5 logic will be change in some places (loader-utils will be update too)

evilebottnawi:
Also can you provide here `https://github.com/postcss/postcss-loader/blob/7e3a95624aeba0634ca9fac178805e052a02df3b/src/index.js#L161` postcss version, future postcss release can have difference ast, i will release new version of `css-loader` in near future, will be great use ast only when `postcss` version is same

michael-ciniawsky:
https://github.com/postcss/postcss-loader/pull/384/commits/6bf774fe24a2eaa87fa5e53065a63af13f51ee9b

evilebottnawi:
@michael-ciniawsky thanks!

michael-ciniawsky:
Should we use the semver major part of the PostCSS version for the ast version and make it a `{Number}` instead?

```diff
{
-  version: '7.0.2'
+  version: 7
}
```

src/index.js
evilebottnawi: > Is there any discussion/documentation about the (planned) changes to the Loader (Runner) API ? (Guessing no)

Only in private messages

In fact, it is only now that work has begun on who is responsible for what, although this should have been done much earlier. And frankly it's bad that we lost you as contributor, I would be glad to see if you would return back. 

> I'm simply sick and tired of working on xyz and at the end of the day it is for nothing concrete because of organisational issues/no feedback and/but on the other hand to top it still every time someone bluntly chimes in and does his/her own thing anyways causing nothing but havoc and burden.

Right now I perform this function and opened for any communication and feedback. 

I know that we have a lot of legacy stuff in many loaders/utils/plugins, which should be removed/refactored, and feel free to do this. Also `webpack-default` is just a repository that helps make the code more consistently. We do not necessarily have to release a huge update (`webpack-default` + breaking change) in one major release - the most important thing is stable code.

evilebottnawi:
@michael-ciniawsky i think better return full version, where necessary I can check only the major version myself, also in theory minor version can change ast too

src/Error.js
```diff
@@ -1,13 +1,13 @@
...
     super(err)
 
-    this.name = 'Syntax Error'
+    const { line, column, reason } = err
+
+    this.name = 'SyntaxError'
 
     this.message = ''
-    this.message += `${this.name} \n\n(${err.line}:${err.column}) ${err.reason}`
+    this.message += `${this.name} \n\n(${line}:${column}) ${reason}`
```
evilebottnawi: You should use `err.reason ? err.reason : err.message` instead `SyntaxName`, not all errors is syntax.

You should check on `line/column` here, because not all error contain `line` and `column`

src/Warning.js
```diff
@@ -0,0 +1,14 @@
+class Warning extends Error {
+  constructor (warning) {
+    super()
+
+    const { text, line, column } = warning
+
+    this.name = 'LoaderWarning'
```
evilebottnawi: If you using OOP for `Error` please use `name` as class name - `Warning`. It is best practice.

src/options.json
```diff
@@ -0,0 +1,61 @@
...
+          "ctx": "should be {Object} (https://github.com/postcss/postcss-loader#context-ctx)",
+          "path": "should be {String} (https://github.com/postcss/postcss-loader#path)"
+        }
+      },
+      "additionalProperties": false
+    },
+    "exec": {
+      "type": "boolean"
+    },
+    "ident": {
```
evilebottnawi: `ident` is property for webpack, we should not check this here, this do webpack already

src/Warning.js
```diff
@@ -0,0 +1,14 @@
...
+    this.message = `\n(${line}:${column}) ${text}\n`
```
evilebottnawi: Use `warning.toString()` here  (https://github.com/postcss/postcss/blob/master/lib/warning.es6#L74)

src/Error.js
michael-ciniawsky: I check for it within the loader already via `err.name = 'CssSyntaxError' ? new SyntaxError(err) : err` so the properties are always defined

src/Warning.js
michael-ciniawsky: I change it

src/options.json
michael-ciniawsky: Since we can't change `addtionalProperties` to `false` due to third parties anyways I will test and remove it

src/Warning.js
michael-ciniawsky: I want consistent formatting with the Error Format and the `toString()` message looks weird imho

**Warning**
```
{{ Name }}

({{ Line }}:{{ Column }}) {{ Text }}

```

```
Warning 

(1:1) I'm a Warning

```

**(Syntax)Error**
```
{{ Name }}

({{ Line }}:{{ Column }}) {{ Message }}

{{ Code Frame }}

```

```
SyntaxError

(1:1) I'm an Error

1 |  
2 >|  .class 
3 |

```

evilebottnawi: @michael-ciniawsky :+1: but you should check `line` and `column` anyway, because not all warnings contain this

michael-ciniawsky: I was not sure about that and took the risk for now, will update

evilebottnawi: @michael-ciniawsky also using own warning text message can lead to loss of information (https://github.com/postcss/postcss/blob/master/lib/warning.es6#L74)

Original
```
postcss-lint:a.css:10:14: Avoid !important
```

You lose `plugin` name, you lose original source map. Be careful postcss `warning` is not error, it is just text message, better use postcss logic for this instead own. There is nothing to worry about that they difference view. Remember - we just wrapper and should provide easy interface. In next version of `postcss` message properties can be changes and it is require modify loader source code, it is not good idea.

test/Warnings.test.js
```diff
@@ -0,0 +1,35 @@
...
+        .split('\n')
+        .slice(1)
+        .join('\n')
+
+      expect(message).toMatchSnapshot()
+    })
+  })
+})
+
+process.on('warning', (warning) => console.warn(warning))
```
michael-ciniawsky: remove

src/index.js
```diff
@@ -137,39 +129,57 @@ module.exports = function loader (css, map, meta) {
...
         }
 
-        if (!meta) meta = {}
+        if (!meta) {
+          meta = {}
+        }
+
+        const ast = {
+          type: 'postcss',
+          version: processor.version[0],
```
michael-ciniawsky: wrong 

```diff
- processor.version[0]
+ processor.version
```

src/Error.js
```diff
@@ -0,0 +1,27 @@
...
+class SyntaxError extends Error {
+  constructor (err) {
+    super(err)
+
+    const { line, column, reason } = err
+
+    this.name = 'SyntaxError'
+
+    this.message = ''
+    this.message += `${this.name} \n\n(${line}:${column}) ${reason}`
```
evilebottnawi: Line and column can be undefined if you don't use source map https://github.com/postcss/postcss/blob/8157278c743564555a22da67da2b5972945462d8/lib/input.es6#L124

evilebottnawi: @michael-ciniawsky also any warning can be `SyntaxError` https://github.com/postcss/postcss/blob/master/lib/warning.es6#L76 and https://github.com/postcss/postcss/blob/8157278c743564555a22da67da2b5972945462d8/lib/node.es6#L84

evilebottnawi: In fact, I do not understand why you asked so much about how they should look if you have not studied the source code before :confused:

src/Warning.js
```diff
@@ -0,0 +1,29 @@
...
+ */
+class Warning extends Error {
+  constructor (warning) {
+    super()
+
+    const { text, line, column } = warning
+
+    this.name = 'Warning'
+
+    if (line && column) {
```
evilebottnawi: Line can be `0`

PR #399, opened 2018-12-15
Ehzhang/error on warnings

ehzhang:
### Notable Changes

👋 
This adds an `emitWarningsAsErrors` option to the postcss-loader config, which would (as the name suggests) emit warnings coming from postcss plugins as errors.

We're interested enforce issues coming as warnings to our build pipelines (and development workflows) and fail. For example, we might have typos in imported variable names `var(--foobar)`, which get emitted as warnings and kind of hidden amongst the webpack output. We'd like to be extra strict for our css warnings 😬

An alternative on our end might be to check `stats.warnings` ourselves, but it would be nice to not have to filter through all other warnings that might not be coming from postcss-loader. Very welcome to feedback/discussion for this - would like to know what you might prefer/if this makes sense as a configuration option on this loader.

Thanks in general for maintaining this library :)

#### `Commit Message Summary (CHANGELOG)`

- Adds an `emitWarningsAsErrors` flag to emit plugin warnings as errors

### Type

- [ ] CI
- [ ] Fix
- [ ] Perf
- [x] Docs
- [ ] Test
- [ ] Style
- [ ] Build
- [ ] Chore
- [x] Feature
- [ ] Refactor

### Issues

> ℹ️ What issue(s) (if any) are closed by your PR?

- Fixes `#1`

### SemVer

- [ ] Fix (:label: Patch)
- [x] Feature (:label: Minor)

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,7 @@ module.exports = {
 |[`config`](#config)|`{Object}`|`undefined`|Set `postcss.config.js` config path && `ctx`|
 |[`plugins`](#plugins)|`{Array\|Function}`|`[]`|Set PostCSS Plugins|
 |[`sourceMap`](#sourcemap)|`{String\|Boolean}`|`false`|Enable Source Maps|
+|[`emitWarningsAsErrors`](#emitwarningsaserrors)|`{Boolean}`|`false`|Emit warnings as errors|
 
 ### `Exec`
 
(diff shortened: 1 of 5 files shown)


ehzhang:
Is it the implementation that isn't a good idea, or the addition of this config value?

I'm not sure what you mean by implementing `result.errors()` - my use case here is to effectively set the "log level" for warnings and escalate them from warnings to errors (and I use `emitError` with the `Warning` class)

evilebottnawi:
@ehzhang it is good practice emit warnings as warnings and errors as errors (all loaders and plugins do it), maybe good idea create issue/PR in webpack and implement this feature in core, not in loader/plugin (otherwise we need implement this option in all loaders/plugins and it is not good)

ehzhang:
Sounds reasonable - kind of the response that I would've expected. I did notice that not many loaders do this (maybe the only exception would be one like ts-loader, that allows setting error levels).

I'll close this - thanks for taking the time to look it over!

PR #425, opened 2020-02-10
feat: re-use postcss ast if passed from previous loader

jquense:
### Notable Changes

Allows reusing a postcss AST if the previous loader passed it. Same logic as css-loader, accepts semver compatible AST versions. It skips doing parser specific logic, since a previous AST makes parsing irrelevant 

### Type

- [ ] CI
- [ ] Fix
- [ ] Perf
- [ ] Docs
- [x] Test
- [ ] Style
- [ ] Build
- [ ] Chore
- [x] Feature
- [ ] Refactor

### SemVer

- [ ] Fix (:label: Patch)
- [x] Feature (:label: Minor)

### Checklist

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,9 +1,12 @@
 const path = require('path')
 
+const { satisfies } = require('semver')
+
 const { getOptions } = require('loader-utils')
 const validateOptions = require('schema-utils')
 
 const postcss = require('postcss')
+const postcssPkg = require('postcss/package.json')
 const postcssrc = require('postcss-load-config')
 
 const Warning = require('./Warning.js')
@@ -31,6 +34,16 @@ function loader (css, map, meta) {
   const file = this.resourcePath
 
   const sourceMap = options.sourceMap
+
+  let ast = null
+  if (
+    meta &&
+    meta.ast &&
+    meta.ast.type === 'postcss' &&
+    satisfies(meta.ast.version, `^${postcssPkg.version}`)
+  ) {
+    ast = meta.ast.root
+  }
 
   Promise.resolve().then(() => {
     const length = Object.keys(options)
@@ -107,7 +120,7 @@ function loader (css, map, meta) {
 
     // Loader Exec (Deprecated)
     // https://webpack.js.org/api/loaders/#deprecated-context-properties
-    if (options.parser === 'postcss-js') {
+    if (!ast && options.parser === 'postcss-js') {
       css = this.exec(css, this.resource)
     }
 
@@ -125,7 +138,7 @@ function loader (css, map, meta) {
 
     // Loader API Exec (Deprecated)
     // https://webpack.js.org/api/loaders/#deprecated-context-properties
-    if (config.exec) {
+    if (!ast && config.exec) {
       css = this.exec(css, this.resource)
     }
 
@@ -138,7 +151,7 @@ function loader (css, map, meta) {
     }
 
     return postcss(plugins)
-      .process(css, options)
+      .process(ast || css, options)
       .then((result) => {
         let { css, map, root, processor, messages } = result
 
@@ -218,6 +231,8 @@ function loader (css, map, meta) {
  * @module postcss-loader
  *
  * @requires path
+ *
+ * @requires semver
  *
  * @requires loader-utils
  * @requires schema-utils
--- a/test/ast-loader.js
+++ b/test/ast-loader.js
@@ -1,0 +1,28 @@
+const postcss = require('postcss')
+const postcssPkg = require('postcss/package.json')
+const semver = require('semver')
+
+const incomingVersion = semver.inc(postcssPkg.version, 'minor')
+
+module.exports = function astLoader (content) {
+  const callback = this.async()
+
+  const { spy = jest.fn() } = this.query
+
+  content = this.exec(content, this.resource)
+
+  postcss()
+    .process(content, { parser: require('postcss-js') })
+    .then(({ css, map, root, messages }) => {
+      const ast = {
+        type: 'postcss',
+        version: incomingVersion
+      }
+
+      Object.defineProperty(ast, 'root', {
+        get: spy.mockReturnValue(root)
+      })
+
+      callback(null, css, map, { ast, messages })
+    })
+}
--- a/test/loader.test.js
+++ b/test/loader.test.js
@@ -1,5 +1,4 @@
 const { webpack } = require('@webpack-utilities/test')
-
 describe('Loader', () => {
   test('Default', () => {
     const config = {
@@ -15,6 +14,34 @@ describe('Loader', () => {
       const { source } = stats.toJson().modules[1]
 
       expect(source).toEqual('module.exports = "a { color: black }\\n"')
+      expect(source).toMatchSnapshot()
+    })
+  })
+
+  test('Loader - Previous AST', () => {
+    const spy = jest.fn()
+    const config = {
+      rules: [
+        {
+          test: /style\.js$/,
+          use: [
+            {
+              loader: require.resolve('../src'),
+              options: { importLoaders: 1 }
+            },
+            {
+              loader: require.resolve('./ast-loader'),
+              options: { spy }
+            }
+          ]
+        }
+      ]
+    }
+
+    return webpack('jss/index.js', config).then((stats) => {
+      const { source } = stats.toJson().modules[1]
+
+      expect(spy).toHaveBeenCalledTimes(1)
       expect(source).toMatchSnapshot()
     })
   })
--- a/package.json
+++ b/package.json
@@ -13,7 +13,8 @@
     "loader-utils": "^1.1.0",
     "postcss": "^7.0.0",
     "postcss-load-config": "^2.0.0",
-    "schema-utils": "^1.0.0"
+    "schema-utils": "^1.0.0",
+    "semver": "^6.3.0"
   },
   "devDependencies": {
     "@webpack-utilities/test": "^1.0.0-alpha.0",
--- a/test/__snapshots__/loader.test.js.snap
+++ b/test/__snapshots__/loader.test.js.snap
@@ -1,3 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Loader Default 1`] = `"module.exports = \\"a { color: black }\\\\n\\""`;
+
+exports[`Loader Loader - Previous AST 1`] = `"module.exports = \\"a {\\\\n    color: yellow\\\\n}\\""`;
+
+exports[`Loader uses previous AST 1`] = `"module.exports = \\"a {\\\\n    color: yellow\\\\n}\\""`;


test/options/exec.test.js
```diff
@@ -25,6 +25,7 @@ describe('Options', () => {
   test('JSS - {String}', () => {
     const config = {
       loader: {
+        test: /style\.js$/,
```
jquense: this test was wrong.

evilebottnawi: Feel free to fix it

jquense: i did!

jquense:
I don't necessarily have strong use-case. Thinking a bit proactively.

you might want to do: `css-loader!postcss-loader!css-module-loader` for autoprefixing. 

Basically the use case is loaders that use postcss but can't be simplified to just a postcss plugin.

evilebottnawi:
Can you fix tests?

jquense:
I can update the snapshots, but I believe these were broken before I made any changes, esp since i haven't touched that code...

jquense:
anything else i can do here?

test/options/sourceMap.test.js
```diff
@@ -47,7 +47,7 @@ describe('Options', () => {
       const { source } = stats.toJson().modules[1]
 
       expect(source).toEqual(
-        'module.exports = "a { color: rgba(255, 0, 0, 1.0) }\\n\\n/*# sourceMappingURL=data:application/json;base64,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbInRlc3QvZml4dHVyZXMvY3NzL3N0eWxlLmNzcyJdLCJuYW1lcyI6W10sIm1hcHBpbmdzIjoiQUFBQSxJQUFJLDJCQUFZLEVBQUUiLCJmaWxlIjoidGVzdC9maXh0dXJlcy9jc3Mvc3R5bGUuY3NzIiwic291cmNlc0NvbnRlbnQiOlsiYSB7IGNvbG9yOiBibGFjayB9XG4iXX0= */"'
+        'module.exports = "a { color: rgba(255, 0, 0, 1.0) }\\n\\n/*# sourceMappingURL=data:application/json;base64,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbInRlc3QvZml4dHVyZXMvY3NzL3N0eWxlLmNzcyJdLCJuYW1lcyI6W10sIm1hcHBpbmdzIjoiQUFBQSxJQUFJLDRCQUFhIiwiZmlsZSI6InRlc3QvZml4dHVyZXMvY3NzL3N0eWxlLmNzcyIsInNvdXJjZXNDb250ZW50IjpbImEgeyBjb2xvcjogYmxhY2sgfVxuIl19 */"'
```
evilebottnawi, line 50: /cc @jquense It is strange, why source map was changed

test/loader.test.js
```diff
@@ -18,4 +17,32 @@ describe('Loader', () => {
       expect(source).toMatchSnapshot()
     })
   })
+
+  test.only('uses previous AST', () => {
```
jquense, line 21:
```suggestion
  test('uses previous AST', () => {
```

test/options/sourceMap.test.js
jquense: i don't know, this was broken before my PR I think: https://github.com/postcss/postcss-loader/pull/425#issuecomment-584228103

evilebottnawi: @jquense maybe postcss have bugs with source maps when we share AST?

jquense: no these tests are failing on master right now

jquense: the source map difference is unrelated to this PR, my guess is it's a difference from node 10 to 12, this repo is really inactive so maybe no one has noticed until now

evilebottnawi: I think it is bug on postcss side when we share AST

evilebottnawi: I faced with same problem in css-loader and put it in my TODO for the next major release

jquense: I really don't think it's a bug with this PR, if I make NO changes to the code on a fresh checkout of postcss-loader/master these two tests fail.

jquense: see: https://github.com/postcss/postcss-loader/pull/427

evilebottnawi: Yep, let's merge, improving source maps for CSS pipeline in my TODO, so let's check and tests all in other PRs

src/index.js
```diff
@@ -1,9 +1,12 @@
 const path = require('path')
 
+const { satisfies } = require('semver')
```
michael-ciniawsky, line 3: I don't really want to add another dependency if possible, do we really need the 'full' `semver` package or would a smaller utility function also solve it e.g 

```js
const SEMVER_RE = /\^?\~?(\d+)\.(\d+).(\d+)/

function semver (version) {
    const [ _, major, minor, patch ] = version.match(SEMVER_RE)
    
    return {
      major,
      minor,
      patch
   }
}

const version = semver(postcssPkg.version)

// ...
```

Please see the comment below for further discussion

src/index.js
```diff
@@ -32,6 +35,16 @@ function loader (css, map, meta) {
 
   const sourceMap = options.sourceMap
 
+  let prevAst = null
+  if (
+    meta &&
+    meta.ast &&
+    meta.ast.type === 'postcss' &&
+    satisfies(meta.ast.version, `^${postcssPkg.version}`)
```
michael-ciniawsky, line 43: What/Which format does `meta.ast.version` currently return? I think the `major` should be enough as it is unlinkely to have AST changes without introducing a BREAKING CHANCE. Needs to be documented somehow in any case.

Assumming to return the `major` version ist schould be sufficient to e.g 

```js
if (
  meta &&
  meta.ast &&
  meta.ast.type === 'postcss' &&
  meta.ast.version === version.major
) ...
```

I also think the `meta` property is always defined, but not 100% sure anymore...

src/index.js
```diff
@@ -32,6 +35,16 @@ function loader (css, map, meta) {
 
   const sourceMap = options.sourceMap
 
+  let prevAst = null
```
michael-ciniawsky, line 38:
```suggestion
  let ast = null
  
```

src/index.js
```diff
@@ -32,6 +35,16 @@ function loader (css, map, meta) {
...
+  if (
+    meta &&
+    meta.ast &&
+    meta.ast.type === 'postcss' &&
+    satisfies(meta.ast.version, `^${postcssPkg.version}`)
+  ) {
+    prevAst = meta.ast.root
```
michael-ciniawsky, line 45:
```suggestion
    ast = meta.ast.root
```

src/index.js
```diff
@@ -107,7 +120,7 @@ function loader (css, map, meta) {
 
     // Loader Exec (Deprecated)
     // https://webpack.js.org/api/loaders/#deprecated-context-properties
-    if (options.parser === 'postcss-js') {
+    if (!prevAst && options.parser === 'postcss-js') {
```
michael-ciniawsky, line 123:
```suggestion
    if (!ast && options.parser === 'postcss-js') {
```

src/index.js
```diff
@@ -125,7 +138,7 @@ function loader (css, map, meta) {
 
     // Loader API Exec (Deprecated)
     // https://webpack.js.org/api/loaders/#deprecated-context-properties
-    if (config.exec) {
+    if (!prevAst && config.exec) {
```
michael-ciniawsky, line 141:
```suggestion
    if (!ast && config.exec) {
```

src/index.js
```diff
@@ -136,9 +149,10 @@ function loader (css, map, meta) {
     if (sourceMap && map) {
       options.map.prev = map
     }
+    const content = prevAst || css
 
     return postcss(plugins)
-      .process(css, options)
+      .process(content, options)
```
michael-ciniawsky, line 155:
```suggestion
      .process(ast || css, options)
```

test/ast-loader.js
```diff
@@ -0,0 +1,28 @@
+const postcss = require('postcss')
```
michael-ciniawsky, line 1: What is this test doing in particular ?

test/loader.test.js
```diff
@@ -18,4 +17,32 @@ describe('Loader', () => {
       expect(source).toMatchSnapshot()
     })
   })
+
+  test('uses previous AST', () => {
```
michael-ciniawsky, line 21:
```suggestion
  test('Loader - Previous AST', () => {
```

src/index.js
jquense: parsing semver with a simple regex is a generally a bad idea, and can fail in unexpected ways. I get not wanting another dependency but css-loader already does this and it's probably more worthwhile to match the behavior esp since the dependency is most likely already installed

evilebottnawi: You regexp is invalid, parsing version using regexp is a bad idea

test/ast-loader.js
evilebottnawi: It is pseudo loader before `postcss-loader` which transmits postcss AST

src/index.js
jquense: it's the full version string from the postcss Processor. hopefully just the major version is enough to determine breaking changes, but ppl and computers are messy and breaking changes  in AST's happen whenever in practice. Yes this a "best guess", but i think it's wise to encode more rather than less information here since we know it. It means that down the road if something does break unexpectedly and we (or consumers of this metadata downstream) need to adjust on a minor or patch version it would be possible, and not require a breaking API change here to add

test/ast-loader.js
jquense: exercising the previous code path! This needs to be a separate file for the ast-loader.test file to use

src/index.js
michael-ciniawsky: I will open an issue in the postcss repo maybe it's possible to add the `version` to e.g the `ast.root` so any third party tool does have to invent it's own way to handle this (not blocking this PR on it) but maybe we are luky...

src/index.js
michael-ciniawsky: Why? I mean I simply not convinced this requires to have all blows and whistles of the semver spec to work and therefore the whole `semver` as a dependecy might be avoidable, but let's not hung ourselves up on this for to long, like mentioned above I will open an issue in the `postcss` repo

test/ast-loader.js
michael-ciniawsky: Understood

michael-ciniawsky:
https://github.com/postcss/postcss/issues/1359

src/index.js
jquense: oo great idea

jquense:
split out test changes and addressed feedback

evilebottnawi:
@jquense friendly ping, still valid?

jquense:
still valid, I haven't had a chance to rebase tho

evilebottnawi:
@jquense can you provide link on loader there it can be used?

jquense:
I don't have a specific case aside from like `css-module-loader`, the other case is another postcss-loader in front of it. I don't have a super strong justification other than "why not? css-loader does it"

evilebottnawi:
@jquense just asked it, because want to know what somebody use it, we will shipped it in the next major release (ETA is current week I think)

jquense:
ya, it's more of an internal optimization that seemed good to do since postcss is often uses as a tooling base for other more specific loader chains

PR #475, opened 2020-09-08
fix: allow inlined sourceMaps when sourceMap or devtool are truthy

alan-agius4:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case
Previously when either the sourceMap option or devtool was truthy
inlining of sourceMaps was not possible.

### Additional Info

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -70,17 +70,16 @@ export default async function loader(content, sourceMap, meta) {
   );
 
   if (useSourceMap) {
-    processOptions.map = { inline: false, annotation: false };
+    if (
+      typeof processOptions.map === 'undefined' ||
+      typeof processOptions.map === 'boolean'
+    ) {
+      processOptions.map = { inline: false, annotation: false };
+    }
 
     if (sourceMap) {
       processOptions.map.prev = normalizeSourceMap(sourceMap, this.context);
     }
-  } else if (sourceMap && typeof processOptions.map !== 'undefined') {
-    if (typeof processOptions.map === 'boolean') {
-      processOptions.map = { inline: true };
-    }
-
-    processOptions.map.prev = sourceMap;
   }
 
   let root;
--- a/test/sourceMap.test.js
+++ b/test/sourceMap.test.js
@@ -428,4 +428,46 @@ describe('"sourceMap" option', () => {
     expect(getWarnings(stats)).toMatchSnapshot('warnings');
     expect(getErrors(stats)).toMatchSnapshot('errors');
   });
+
+  it('should generate inline source maps with "true" value nd the "devtool" with "false", but the "postcssOptions.map" has inline map values', async () => {
+    const compiler = getCompiler(
+      './css/index.js',
+      {
+        sourceMap: true,
+        postcssOptions: {
+          sourceMap: true,
+          map: {
+            inline: true,
+            annotation: false,
+          },
+        },
+      },
+      {
+        devtool: false,
+      }
+    );
+    const stats = await compile(compiler);
+    const { css } = getCodeFromBundle('style.css', stats);
+    expect(css).toMatchSnapshot('css');
+  });
+
+  it('should generate inline source maps when value is not specified and the "devtool" with "source-map", but the "postcssOptions.map" has inline map values', async () => {
+    const compiler = getCompiler(
+      './css/index.js',
+      {
+        postcssOptions: {
+          map: {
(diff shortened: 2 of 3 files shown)


src/index.js
```diff
@@ -70,17 +70,18 @@ export default async function loader(content, sourceMap, meta) {
...
+      typeof processOptions.map === 'boolean'
+    ) {
+      processOptions.map = { inline: false, annotation: false };
+    }
 
-    if (sourceMap) {
+    if (sourceMap && !processOptions.map.inline) {
       processOptions.map.prev = normalizeSourceMap(sourceMap, this.context);
+    } else {
+      processOptions.map.prev = sourceMap;
```
evilebottnawi, line 83: When we set `sourceMap: true` we should always normalize source map, otherwise you will get unstable `contenthash` on different platforms due absolute paths in sources from other loaders

alan-agius4: Updated

test/sourceMap.test.js
```diff
@@ -428,4 +428,46 @@ describe('"sourceMap" option', () => {
...
     expect(getErrors(stats)).toMatchSnapshot('errors');
   });
+
+  it('should generate inline source maps with "true" value nd the "devtool" with "false", but the "postcssOptions.map" has inline map values', async () => {
+    const compiler = getCompiler(
+      './css/index.js',
+      {
+        sourceMap: true,
+        postcssOptions: {
+          sourceMap: true,
```
evilebottnawi, line 438: No need :+1:

src/index.js
```diff
@@ -70,17 +70,16 @@ export default async function loader(content, sourceMap, meta) {
   );
 
   if (useSourceMap) {
-    processOptions.map = { inline: false, annotation: false };
+    if (
+      typeof processOptions.map === 'undefined' ||
+      typeof processOptions.map === 'boolean'
```
evilebottnawi, line 75: `processOptions.map` means as the same `postcss(plugins).process(content, { from, map: true })`, there is not docs for this, but it should generate inline map as I remember

src/index.js
```diff
@@ -70,17 +70,16 @@ export default async function loader(content, sourceMap, meta) {
...
 
     if (sourceMap) {
       processOptions.map.prev = normalizeSourceMap(sourceMap, this.context);
     }
-  } else if (sourceMap && typeof processOptions.map !== 'undefined') {
-    if (typeof processOptions.map === 'boolean') {
-      processOptions.map = { inline: true };
-    }
-
-    processOptions.map.prev = sourceMap;
```
evilebottnawi, line 83: Also it is breaking change, now you need always to set `sourceMap: true` if you need source maps, but it is invalid in some cases I don't need webpack source maps, I need only postcss source maps (for example generate library with css and relative urls in sources)

src/index.js
alan-agius4: Actually it should generate external sourcemaps

https://github.com/postcss/postcss/blob/master/docs/source-maps.md

src/index.js
alan-agius4: Maybe I can amend the logic to check for either `useSourceMaps` or `postcssOptions.map`?

src/index.js
evilebottnawi: >To generate a new source map with the default options, simply set map: true. This will generate an inline source map that contains the source content. If you don’t want the map inlined, you can set map.inline: false.

src/index.js
evilebottnawi: What do you want to achieve? I don't understand what you are trying to do, if you want to generate inline source map just set `postcssOptions.map: true`

alan-agius4: At the moment if devtool is set you cannot “inline” a sourcemap.

evilebottnawi:
@alan-agius4 After this PR https://github.com/webpack-contrib/postcss-loader/pull/476 - you should not use `sourceMap: true` and `postcssOptions.map` together, this makes no sense, using the `postcssOptions.map`` option enable source maps generation

alan-agius4:
Thanks

evilebottnawi:
@alan-agius4 feel free to feedback

alan-agius4:
@evilebottnawi, the latest patch version fixes the problem. Thanks

Issue #510, opened 2021-01-14
Accept a custom postcss in the config?


phated:
- Operating System: Mac
- Node Version: ^14
- NPM Version: ^6
- webpack Version: ^4
- postcss-loader Version: 4.1.0

### Feature Proposal

I'd like to pass the `postcss()` function as an argument to this loader. If the `postcss` function isn't passed, it can fallback to loading from the peerDependency.

I'm happy to contribute this feature if you think it would be helpful!

### Feature Use Case

`postcss-loader` is embedded deep into the core of Storybook and v3 implicitly caused a dependency on PostCSS 7. That is fine, because this loader can be upgraded and a dependency for `^7.0.35` can be added. However, some consumers want PostCSS 8, but it is a breaking change to bump that internal dependency.

If the `postcss()` function could be specified, it could be exposed as a configuration option that allows people to opt-in to PostCSS 8 support but would fallback to the v7 dependency otherwise.

alexander-akait:
Sorry, out of scope, it will only confuse new developers, and people will start using it without understanding why it is needed, you can validate version of postcss on your side and throw error/warning, you can even do it inside configuration

phated:
Thanks for your feedback. We will be able to do your suggestion in the future by requiring users to add postcss as a peerDep, but that is a breaking change to our current users.

We will likely have to use a fork of this loader until we are able to make a breaking change.

alexander-akait:
Fixed in master, release will be today

PR #511, opened 2021-01-15
Allow tooling authors to provide custom postcss factory through options

phated:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Ref #510 

I know @alexander-akait said it is out-of-scope, but I needed to write the code anyway. So I figured I'd make the code suitable to upstream and submit it anyway—under the assumption this will be closed.

I'll be publishing my fork as `@phated/postcss-loader` for our use in Storybook until we can make the 7.0 breaking release—at which time, we'll be able to require users to rely on peerDeps and switch back to this upstream project.

Cheers! 🚀

### DIFF
--- a/README.md
+++ b/README.md
@@ -593,6 +593,37 @@ module.exports = {
 };
 ```
 
+### `implementation`
+
+Type: `Function`
+
+The special `implementation` option determines which implementation of PostCSS to use. Overrides the locally installed `peerDependency` version of `postcss`.
+
+**This option is only really useful for downstream tooling authors to ease the PostCSS 7-to-8 transition.**
+
+**webpack.config.js**
+
+```js
+module.exports = {
+  module: {
+    rules: [
+      {
+        test: /\.css$/i,
+        use: [
+          { loader: "style-loader" },
+          { loader: "css-loader" },
+          {
+            loader: "postcss-loader",
+            options: { implementation: require("postcss") },
+          },
+          { loader: "sass-loader" },
+        ],
+      },
+    ],
+  },
+};
+```
+
 ## Examples
 
 ### SugarSS
--- a/src/index.js
+++ b/src/index.js
@@ -46,6 +46,8 @@ export default async function loader(content, sourceMap, meta) {
       ? true
       : options.postcssOptions.config;
 
+  const postcssFactory = options.implementation || postcss;
+
   let loadedConfig;
 
   if (configOption) {
@@ -105,7 +107,10 @@ export default async function loader(content, sourceMap, meta) {
   let result;
 
   try {
-    result = await postcss(plugins).process(root || content, processOptions);
+    result = await postcssFactory(plugins).process(
+      root || content,
+      processOptions
+    );
   } catch (error) {
     if (error.file) {
       this.addDependency(error.file);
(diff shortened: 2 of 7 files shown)


alexander-akait (commented):
Can you provide example of usage? Do you want to use `postcss@7` or `postcss@8`? Why not validate it on boilerplate side? In theory I can add this as the migration options (and when we drop postcss@7 support we remove it)

phated:
Currently, the way a user consumes Storybook looks (more-or-less) like:

```
/users-project-root
├─┬ node_modules
│ └─┬ storybook
│   ├── postcss-loader
│   └── postcss@^7
└── storybook-config-file
```

The internally-included `postcss@^7` dependency cannot be removed before the 7.0 released, and users are using it implicitly.

However, a new option can be exposed that allows users to specify a custom postcss factory that has priority over the implicit dependency using this feature. So they would do:

```
/users-project-root
├─┬ node_modules
│ ├─┬ storybook
│ │ ├── postcss-loader
│ │ └── postcss@^7
│ └── postcss@^8 <-- add postcss 8 as top-level dep
└── storybook-config-file <-- add require("postcss")
```

alexander-akait (commented):
Okay, let's rename `customPostcss` to `implementation`, we have the same option in `sass-loader` (https://github.com/webpack-contrib/sass-loader/blob/master/src/options.json#L4)

phated:
@alexander-akait great thought! I've updated to be more consistent with sass-loader. Do you have any recommendations for getting the snapshots to succeed consistently across node versions?

test/cjs.test.js
```diff
@@ -1,3 +1,4 @@
+// eslint-disable-next-line import/no-named-as-default, import/no-named-as-default-member
```
alexander-akait, line 1: I think it is global eslint problem, we don't have this rules

src/index.js
```diff
@@ -1,7 +1,7 @@
 import { getOptions } from "loader-utils";
 import { validate } from "schema-utils";
 
-import postcss from "postcss";
+import ourPostcss from "postcss";
```
alexander-akait, line 4: Let's return to `postcss`

README.md
```diff
@@ -593,6 +593,38 @@ module.exports = {
 };
 ```
 
+### `implementation`
+
+Type: `Function`
+Default: `undefined`
```
alexander-akait, line 599: `require('postcss')`

alexander-akait:
Also can you accept CLA?

phated:
I think the CLA app is bugged

Screen Shot 2021-01-20 at 10 35 00 AM

src/index.js
phated: The problem I ran into was naming the local variable `postcss` too. I can change that and put this back to `postcss`

phated:
@alexander-akait I addressed your comments and squashed the commits. I removed my package-lock updates and other things that made the tests pass locally for me, I think those should be fixed outside this feature.

alexander-akait (approved):

alexander-akait:
Thanks

PR #527, opened 2021-05-15
feat: check postcss versions to avoid using PostCSS 7

### LINKED ISSUE #507: Check postcss versions to avoid using PostCSS 7
Many users report the same problem with PostCSS 7 → 8 migration.

1. They updated `postcss-loader`
2. They ignored `postcss` peer dependency warning
3. npm takes old `postcss` 7 from some deep dependency
4. As a result they have an error about PostCSS incompatibility

Of course, we can blame users. But this is a popular error.

In `postcss-cli` we check `postcss().version` and show `Add postcss to package.json` error message. Should we do something like this in `postcss-loader`?

shlyk:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Added code that helps the developer understand that it is necessary to use an explicit version of PostCSS. Closes #507.

### Breaking Changes

No breaking changes.

### Additional Info

cc @ai


ai:
LGTM

@alexander-akait can you start CI?

src/index.js
```diff
@@ -38,6 +41,20 @@ export default async function loader(content, sourceMap, meta) {
 
   const postcssFactory = options.implementation || postcss;
 
+  // Check postcss versions to avoid using PostCSS 7
+  if (!isCheckedPostCSSVersion && postcss().version.startsWith("7.")) {
+    isCheckedPostCSSVersion = true;
+    const pkg = readPackageJson();
```
alexander-akait, line 47: It will add one more fs call for each `css` file, reduce perf very well

ai: As I understand, `fs` call will be only once and only for `postcss 7`

alexander-akait: We can solve it on two ways:
- webpack already reads `package.json`, need look where we store it
- using `this.fs`, so each call will be cacheable

alexander-akait: But it is invalid, in monorepos you can have multiple packages with different versions of postcss

shlyk: OK, one minute, I fix it.

ai:
Closes https://github.com/webpack-contrib/postcss-loader/issues/507

src/index.js
shlyk: Perhaps it is better to import the package.the json file at the beginning. So it will be cached: 'const pkg = require(resolve(process. cmd ()," package.json"));`

Your opinion?

alexander-akait:
We will positive false/false positive for monorepos and will spam warnings, it is bad

ai:
Anyway we will run this check only for PostCSS 7 case, which will be rare case.

Should we add deep search and check for `package.json`?

alexander-akait:
`15638766` downloads by week https://www.npmjs.com/package/postcss, it is big number, can't say how many uses potcss-loader, but I think it is big number too. 

> Should we add deep search and check for package.json? Another option is an environment variable to disable check in rare cases.

To be honest, this is not an easy task, you can see the algorithm in npm/yarn/pnpm and found it is not easy. Also it is not make sense for yarn PnP/pnpm (will be failed by default) and npm v7 (automatically install peer deps)

src/index.js
shlyk: @alexander-akait can you show examples of such repositories? I'm talking about the mono repository.

ai:
> 15638766 downloads by week https://www.npmjs.com/package/postcss, it is big number, can't say how many uses potcss-loader, but I think it is big number too.

A lot of PostCSS 7 downloads is coming from plugins. The real number is much smaller.

Right now using PostCSS 7 is insecure because of vulnerability. It forced migration. 

> To be honest, this is not an easy task, you can see the algorithm in npm/yarn/pnpm and found it is not easy.

We do not need to cover all cases here. We just need to prevent popular error. Edge cases can be fixed by:

1. We can have an environment variable to disable check
2. We can use warning instead of error.

alexander-akait:
> I think npm 7 do not install peer dependencies for old dependencies on npm 6→7 migration. It will help only for new project or during installing new postcss-loader.

https://github.blog/2021-02-02-npm-7-is-now-generally-available/#peer-dependencies

ai:
> Since many packages in the ecosystem have come to rely on loose peer dependencies resolutions, npm 7 will print a warning and work around most peer conflicts that exist deep within the package tree, since you can’t fix those anyway.

Yeap. As I understand, npm 7 will not fix old missed peer dependencies.

shlyk:
Please help me understand what to do next. Add a warning and an environment variable? Or do you want to do this task more thoroughly in another PR?

alexander-akait:
> Please help me understand what to do next. Add a warning and an environment variable? Or do you want to do this task more thoroughly in another PR?

I understand your excitement, but it's okay, we just need to discuss a little more and as soon as we come to a decision I immediately ping you

alexander-akait:
> Yeap. As I understand, npm 7 will not fix old missed peer dependencies.

npm 7 install them by default, no matter whether it is old or new

Anyway the main problem is not npm/yarn/pnpm, searching package.json is the main problem. 

I think we should use, simplified, but covered most of cases:
```js
const cwd = process.cwd();
let dir = cwd;
for (;;) {
    try {
        // `this` - is loader context
        if (this.fs.statSync(path.join(dir, 'package.json')).isFile()) break;
    } catch (e) {}
    const parent = path.dirname(dir);
    if (dir === parent) {
      dir = undefined;
      break;
    }
    dir = parent;
}
```

It's just that from my experience these things usually create more problems than they solve... Package manager output is the place where you **need** read warnings/errors. All of these workarounds are trying to train the developer to be attentive, but if you do not read errors from your tty, you may need to learn more.. Just my opinion.

ai:
yarn output is full on warnings about deprecated nested dependencies, which you can't fix. They hide important warning which you can fix.

We know that people ignore this warning. Speaking about “the right way” will not solve the problem.

Another way to fix the problem, is to catch errors from PostCSS and show warning about missed dependency before re-throwing the error.

alexander-akait:
> Another way to fix the problem, is to catch errors from PostCSS and show warning about missed dependency before re-throwing the error.

Sounds very good, I like it

ai:
@shlyk a new plan:

1. Wrap `postcss.process` can in `try catch`
2. On error from `postcss` we are starting old check
3. Instead of throwing error, we print warning

shlyk:
> @shlyk a new plan:
> 
> 1. Wrap `postcss.process` can in `try catch`
> 2. On error from `postcss` we are starting old check
> 3. Instead of throwing error, we print warning

OK

shlyk:
@ai and @alexander-akait, please check it out.

src/utils.js
```diff
@@ -408,10 +409,20 @@ function normalizeSourceMapAfterPostcss(map, resourceContext) {
   return newMap;
 }
 
+function readPackageJson(readFileSync) {
+  return JSON.parse(
+    (readFileSync || fs.readFileSync)(
```
ai, line 414: You need to check that file exists first.

Otherwise it will throw `Can’t read file` error if `package.json` was in a different place.

src/index.js
```diff
@@ -102,6 +104,22 @@ export default async function loader(content, sourceMap, meta) {
       processOptions
     );
   } catch (error) {
+    // Check postcss versions to avoid using PostCSS 7.
+    if (
+      !hasExplicitDependencyOnPostCSS &&
+      postcssFactory().version.startsWith("7.")
+    ) {
+      // For caching reasons, we use the readFileSync function from the context, not the function from the `fs` module.
+      const pkg = readPackageJson(this.fs.readFileSync);
```
ai, line 113: You need to check that `pacjage.json` exists first (print nothing on no file).

Otherwise, it will throw Can’t read file error if package.json was in a different place.

shlyk: Thanks, done.

src/index.js
```diff
@@ -102,6 +109,25 @@ export default async function loader(content, sourceMap, meta) {
       processOptions
     );
   } catch (error) {
+    // Check postcss versions to avoid using PostCSS 7.
+    // For caching reasons, we use the readFileSync and existsSync functions from the context,
+    // not the functions from the `fs` module.
+    if (
+      !hasExplicitDependencyOnPostCSS &&
+      isFileExists(PACKAGE_JSON_PATH, this.fs.existsSync) &&
```
ai, line 117: Run `postcssFactory().version.startsWith("7.")` before `isFileExists` to avoid unnecessary file check for PostCSS 8 users

src/utils.js
```diff
@@ -408,10 +409,24 @@ function normalizeSourceMapAfterPostcss(map, resourceContext) {
   return newMap;
 }
 
+function isFileExists(filePath, existsSync = fs.existsSync) {
```
ai, line 412: Do we need default value for any cases?

shlyk: You're right, I removed the default value. They don't seem to be needed here.

src/index.js
shlyk: I also changed it to call `statSync(...).isFile()` instead of `existsSync()`, because `existsSync()` is not available in this version of Webpack.

ai:
@alexander-akait we are ready for your review

src/index.js
```diff
@@ -11,8 +13,14 @@ import {
   exec,
   normalizeSourceMap,
   normalizeSourceMapAfterPostcss,
+  parsePackageJson,
+  isFileExists,
 } from "./utils";
 
+let hasExplicitDependencyOnPostCSS = false;
+
+const PACKAGE_JSON_PATH = path.resolve(process.cwd(), "package.json");
```
alexander-akait, line 22: cwd can be `src` (i.e. `src/package.json`), but it is invalid I provided example above with recusing searching

shlyk: it would be possible to use iterators to search for a file (from similar solutions), but it seems that this is not essential here. I took your solution, especially since there is a similar solution in the Webpack source code.

shlyk: I also use `null and `undefined` to cache calls. For some, this may seem implicit. I can try to make it more explicit, if necessary.

shlyk:
@alexander-akait and @ai, are we doing something extra? Any other questions?

ai:
LGTM. @alexander-akait let’s merge =^_^=

alexander-akait:
I'm a little busy now, tomorrow I will merge and do release :+1:

alexander-akait:
@shlyk Can you accept CLA (link in github actions checks)?

shlyk:
@alexander-akait done.

alexander-akait:
You need to use the same email in commits, can you add it?

shlyk:
@alexander-akait done.

shlyk:
@alexander-akait Э, it looks like something needs to be done. Is there anything I can do to help?

alexander-akait:
@shlyk You still have commits without email (look at your commits above), can you fix it?

PR #568, opened 2022-02-09
docs(readme): added types in readme

harish-sethuraman:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [x] **metadata update**

### Motivation / Use-Case
Added proper type definitions in readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -650,8 +650,10 @@ module.exports = {
 Type:
 
 ```ts
-type implementation = require('postcss') | string
+type implementation = object;
 ```
+
+type of `implementation` should be the same as [postcss.d.ts](https://github.com/postcss/postcss/blob/main/lib/postcss.d.ts)
 
 Default: `postcss`
 
--- a/README.md
+++ b/README.md
@@ -142,18 +142,22 @@ And run `webpack` via your preferred method.
 
 ## Options
 
-|                Name                 |         Type         |                Default                | Description                                  |
-| :---------------------------------: | :------------------: | :-----------------------------------: | :------------------------------------------- |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -612,14 +647,19 @@ module.exports = {
 
 ### `implementation`
 
-Type: `Function | String`
+Type:
+
+```ts
+type implementation = require('postcss') | string
```
alexander-akait, line 653: It can be function, let's improve this, `require('postcss')` is default value

harish-sethuraman: `require('postcss')` exports a type file called `postcss.d.ts`. That is why I gave `require('postcss')`

harish-sethuraman: 🤔  There is no type for package? Shall I leave it generic as `object`?

alexander-akait: I think we can do it, just add comment about it should be have the same API as postcss

harish-sethuraman, line 653:
```suggestion
type implementation = object
```

README.md
```diff
@@ -612,14 +647,19 @@ module.exports = {
...
+```
+
```
harish-sethuraman, line 655:
```suggestion
type of implementation should be the same as [postcss.d.ts](https://github.com/postcss/postcss/blob/main/lib/postcss.d.ts)
```

harish-sethuraman: Is this okay ?

alexander-akait: :+1:

Issue #638, opened 2023-04-03
Webpack out of memory error with postcss-loader 7.2.x


strmer15:
### Bug report

Every production webpack build is failing with an out of memory error once I update to `postcss-loader` 7.2.x. As soon as I downgrade back to 7.1.x, it works again.

### Actual Behavior

```
<--- Last few GCs --->

[90131:0x7242530]    65732 ms: Scavenge (reduce) 3798.4 (3990.6) -> 3798.3 (3991.9) MB, 4.9 / 0.0 ms  (average mu = 0.926, current mu = 0.892) allocation failure; 
[90131:0x7242530]    67086 ms: Mark-sweep (reduce) 4144.9 (4337.8) -> 3968.3 (4149.8) MB, 645.9 / 0.1 ms  (+ 102.3 ms in 1953 steps since start of marking, biggest step 0.3 ms, walltime since start of marking 2153 ms) (average mu = 0.846, current mu = 0.7

<--- JS stacktrace --->

FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
 1: 0xb7b3e0 node::Abort() [node]
 2: 0xa8c8aa  [node]
 3: 0xd69100 v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [node]
 4: 0xd694a7 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [node]
 5: 0xf46ba5  [node]
 6: 0xf5908d v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReason, v8::GCCallbackFlags) [node]
 7: 0xf5bff5 v8::internal::Heap::HandleGCRequest() [node]
 8: 0xeda14f v8::internal::StackGuard::HandleInterrupts() [node]
 9: 0x12da75f v8::internal::Runtime_StackGuardWithGap(int, unsigned long*, v8::internal::Isolate*) [node]
10: 0x1707b79  [node]
Aborted (core dumped)
```

### Expected Behavior

It should not be throwing out of memory errors and core dumps.

### How Do We Reproduce?

Run a webpack build with the latest `postcss-loader`? All the webpack builds in our private repos fail with the latest version of `postcss-loader`, but I can't post those configurations publicly.

### Please paste the results of `npx webpack info` here, and mention other relevant information

```
npx webpack info

  System:
[...]

PR #641, opened 2023-04-04
fix: duplicate init typescriptLoader

Vrainy:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Additional Info
Each time the `typescriptLoader` is executed, ts-node will execute `registerExtension()` once.

Multiple executions of the `typescriptLoader` will get
`require.extensions[ext] = wrapper(wrapper(wrapper(wrapper(wrapper(... wrapper(originalHandler)))))`.
[ts-node source code](https://github.com/TypeStrong/ts-node/blob/71dcfd7b813a4178d89562bb4c28d7d6579fae4f/src/index.ts#L1532)

This will cause the out of memory when `require.extensions[ext]` is executed. #638

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -34,6 +34,8 @@ function exec(code, loaderContext) {
   return module.exports;
 }
 
+let tsloader;
+
 async function loadConfig(loaderContext, config, postcssOptions) {
   const searchPath =
     typeof config === "string"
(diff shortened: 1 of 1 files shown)


alexander-akait (approved):
Oh, good catch, souds like node-ts has memory leak...

PR #658, opened 2023-06-09
fix(perf): avoid using `klona` for postcss options

alexander-akait:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Do not use `klona` - better perf and less deps

### Breaking Changes

No

### Additional Info

No

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -343,9 +343,6 @@ async function getPostcssOptions(
     );
   }
 
-  // No need them for processOptions
-  const { plugins: __plugins, ...optionsFromConfig } = processOptionsFromConfig;
-
   const processOptionsFromOptions = { ...normalizedPostcssOptions };
 
   if (processOptionsFromOptions.from) {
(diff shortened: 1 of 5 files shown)


src/utils.js
```diff
@@ -185,11 +184,9 @@ async function loadConfig(loaderContext, config, postcssOptions) {
       options: postcssOptions || {},
     };
 
-    result.config = result.config(api);
+    return { ...result, config: result.config(api) };
```
lionel-bijaoui, line 187: Very uninformed questions, but this is not the same as a deep clone :
- Will it not create issue ? 
- Was a deep clone here an overkill in the first place ?

Thank you for your time ! You are a pillar of the Webpack ecosystem !

lionel-bijaoui: Ok thank you so mush for the details and your time. Have a nice day !

PR #659, opened 2023-06-09
chore(deps): update before release

alexander-akait:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [x] **metadata update**

### Motivation / Use-Case

regenerate lock file and validate latest version before release

### Breaking Changes

No

### Additional Info

No

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "7.3.2",
       "license": "MIT",
       "dependencies": {
-        "cosmiconfig": "^8.1.3",
+        "cosmiconfig": "^8.2.0",
         "jiti": "^1.18.2",
         "semver": "^7.3.8"
       },
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,193 +1,303 @@
 {
   "name": "postcss-loader",
   "version": "7.3.2",
-  "lockfileVersion": 1,
+  "lockfileVersion": 3,
   "requires": true,
-  "dependencies": {
-    "@ampproject/remapping": {
-      "version": "2.2.0",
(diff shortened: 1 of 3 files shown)


alexander-akait:
@d-fischer there is a breaking change with `.mjs` implementation

Example:
```js
import { cosmiconfig, defaultLoaders } from "cosmiconfig";

console.log(defaultLoaders[".js"]);
```

Before it was **sync** function https://github.com/cosmiconfig/cosmiconfig/blob/v8.0.0/src/loaders.ts#L7, now it is async, I am afraid you will get a lot of issue about it soon, I think not many have been updated yet, so everything is fine for now.

I can patch it here, but a lot of CLI tools use postcss-loader and can't easy update it.

alexander-akait:
Perhaps we should have introduced this improvement under a new major release, but it is too late to change it.

d-fischer:
Just wondering - as the containing function is already async, why didn't you just add an `await`?

I agree that multiple people missed that this is breaking, and should have been a major release, but reverting now would by itself be another breaking change.

alexander-akait:
@d-fischer 

> Just wondering - as the containing function is already async, why didn't you just add an await?

Due to jest and Node.js, we have a custom implementation for loading to avoid memory leak, so I want to just use `import-refresh` with `require` here and do not try to `import(...)` if it was failed, it is faster and due `new Function(...)` Node.js remove it from cache after gc

> I agree that multiple people including myself missed that this is breaking, and should have been part of a major release, but reverting now would by itself be another breaking change.

yeah, I agree, reverting this will be a problem more, maybe we should add notes in changelog about it, so it wasn't a surprise

PR #672, opened 2023-10-12
Update package.json

basurohit77:
It has dependency of "postcss": "^8.4.29", which is <8.4.31 according to 
https://github.com/advisories/GHSA-7fh5-64p2-3v2j

This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "memfs": "^3.4.13",
     "midas": "^2.0.3",
     "npm-run-all": "^4.1.5",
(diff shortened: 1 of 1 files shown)


alexander-akait:
You can update your deps locally, don't need to update them here, because we use `^` to specify version

PR #680, opened 2024-01-29
test: sourcemaps with an actual plugin

romainmenke:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

PostCSS has two modes:
- `LazyResult`
- `NoWorkResult`

`NoWorkResult` is an optimized path that is taken if no plugins, parsers, ... are provided.
The assumption here is that the user has a bad config and doesn't actually want to use PostCSS.

`LazyResult` is PostCSS actually doing things.

The sourcemaps generated for both modes are radically different.

This changes adds a noop plugin to trigger the `LazyResult` mode.

### Breaking Changes

N/A

### Additional Info

I don't know if this might surface subtle bugs here.
Main purpose is to have better test coverage.

I am trying to pinpoint the cause of this issue : https://github.com/postcss/postcss/issues/1914

### DIFF
--- a/test/__snapshots__/sourceMap.test.js.snap
+++ b/test/__snapshots__/sourceMap.test.js.snap
@@ -108,7 +108,7 @@ exports[`"sourceMap" option should generate source maps using the "postcssOption
 "a {
   color: coral;
 }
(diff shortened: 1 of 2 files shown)


romainmenke:
Changed my mind, better not to change these tests unless there is an actual issue.

